PR #8, opened 2021-01-14
Reqs adjustment

markurtz:
finish implementation, rearchitected such that:
- flattened structure for objects
- all files are downloadable on their own
- name and pathing scheme changes for latest repo iteration
- convenience methods added to Model class for search an download
- convenience constructors for common models added
- data loader and sets based on numpy implementations added


kevinaer (dismissed):
left a few minor comments but outside of that LGTM, everything seems to be working

bfineran (commented):
Looks good to me overall, a few nits and overall comments:

* I like the model download wrappers, my worry is that especially right now for any given model there are a lot of possible parameter combinations for each wrapper, but most models will only have 1-4 combination of parameters that actually exist in the zoo.  Maybe for the next release we could have some sort of structure over these wrappers that lets users easily see which versions of the model are downloadable
* Can we pick the ~4 minimal/most important functions and propagate them up to the top level `__init__.py` file? This may already be the case since there's a lot of file movement here and hard to track. But it would be nice to be able to do `from sparsezoo import Model, search_models, ModelArgs, download_file` etc for the most common features
* mentioned it in a file comment, but it would be good if we could eventually standardize between `.format` and f-strings.

kevinaer (approved):
LGTM

bfineran (approved):
LGTM

PR #24, opened 2021-01-30
Comprehensive SparseZoo tests

kevinaer:
Added tests capturing all models currently deployed. By default only a small subset run but which model tests run can be configured by setting:
- `SPARSEZOO_TEST_ALL_{MODEL}` to run all tests for a specific model architecture
- `SPARSEZOO_TEST_ALL_{SUB_DOMAIN}` to run all tests under a specific sub domain
- `SPARSEZOO_TEST_ALL_MODELS` to run all tests


bfineran (dismissed):
LGTM. Is the default behavior that users need to enable the tests they want to run? Or will all run by default?

markurtz (changes requested):
Testing the helpers is good, but we aren't planning on having helper functions for every model in the zoo. We should have a generic test that uses search models and validates all of the models returned in the search. This function would be an extensive test.

Also, I'm trying to move us away from environment variables for controlling the tests and instead go through pytests's setup for ignoring files and directories. In this way there isn't any potentially hidden state to the user when running make test and it's a bit easier to document. See the sparseml makefile for the latest on that

markurtz (dismissed):

markurtz (dismissed):

bfineran (approved):

markurtz (approved):

PR #37, opened 2021-02-13
Zoo functions for loading recipes and base checkpoint weights from stubs

bfineran:
intended use cases:
* load recipe from stub instead of local file. (i.e. `ScheduledModifierManager.from_yaml("zoo:/model/stub/path")`)
* get the base weights for a recipe ("base-none" weights for a user trying to reproduce pruning results, and the final model optimized weights for transfer learning)

### DIFF
--- a/src/sparsezoo/models/zoo.py
+++ b/src/sparsezoo/models/zoo.py
@@ -20,11 +20,18 @@ Code for managing the search and creation of sparsezoo Model and Recipe objects
 from typing import List, Union
 
 from sparsezoo.objects.model import Model
-from sparsezoo.objects.optimization_recipe import OptimizationRecipe
+from sparsezoo.objects.optimization_recipe import (
+    OptimizationRecipe,
+    OptimizationRecipeTypes,
+)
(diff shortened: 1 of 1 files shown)


markurtz (approved):

src/sparsezoo/models/zoo.py
```diff
@@ -20,13 +20,20 @@
...
+    OptimizationRecipeTypes,
+)
 from sparsezoo.requests import ModelArgs, download_get_request, search_get_request
 
 
 __all__ = ["Zoo"]
 
 
+# optional prefix for stubs
+_ZOO_STUB_PREFIX = "zoo:"
```
markurtz, line 34: Small thing, can we declare this as public and export it so other integrations can use it explicitly?

markurtz (changes requested):

bfineran:
updated to support query strings as discussed in https://github.com/neuralmagic/sparseml/pull/72
i.e. `zoo:model/stub/path?recipe_type=transfer`

markurtz (approved):

mgoin (approved):

PR #39, opened 2021-02-23
base weight loading for pytorch sparse quant models

bfineran:


bfineran:
blocks https://github.com/neuralmagic/sparseml/pull/81

markurtz (changes requested):
I think we should likely clean up the flow in this function. There are a few edge cases in here that can get by right now; specifically: passing in an invalid recipe type, returning checkpoint files if recipe type is not transfer learning, etc.

I'd recommend we reorganize the logic to gathering and downloading the framework files from the correct base model based on if it's transfer learning or the original sparsification run (and failing if the recipe type doesn't match one of these; None would match to original), and then restricting those files for the ckpt or end framework file based on what was requested

src/sparsezoo/models/zoo.py
```diff
@@ -541,12 +541,28 @@ def download_recipe_base_framework_files(
...
-        if (
-            recipe_type in args
-            and args["recipe_type"] == OptimizationRecipeTypes.TRANSFER_LEARN.value
-        ):
+        # check recipe type, default to original, and validate
+        recipe_type = args.get("recipe_type", OptimizationRecipeTypes.ORIGINAL.value)
+        valid_recipe_types = list(map(lambda typ: typ.value, OptimizationRecipeTypes))
+        if recipe_type not in valid_recipe_types:
+            raise ValueError(
+                f"Invalid reicpe_type {recipe_type}. "
```
mgoin, line 549: *recipe_type

markurtz (approved):

mgoin (approved):

PR #44, opened 2021-02-24
Log Level Changes

kevinaer:
- Logging level to warn
- changed credential log to debug
- increased credential duration to 30 days


markurtz (approved):

bfineran (approved):
LGTM. verified that credentials path is not logged by default

PR #54, opened 2021-03-26
Test Mode and Search Fix

kevinaer:
- Added environment flag to disable increment download count when downloading a model.
- Added a fix to search where release version was query was not properly added to URL


Makefile
```diff
@@ -53,6 +54,7 @@ style:
 # run tests for the repo
 test:
 	@echo "Running python tests";
+	export SPARSEZOO_TEST_MODE=$(SPARSEZOO_TEST_MODE);
```
bfineran, line 57: do we need to unset this after testing?

kevinaer: I actually set the variable at the top of the makefile so it is only within the context of the makefile. This line isn't necessary but I forgot to delete it.

bfineran: oh nice, I didn't realize the Makefile vars would be passed to the environment.  that's neat.

bfineran (approved):

markurtz (approved):

PR #55, opened 2021-03-26
Get Model Path Intergation

kevinaer:
- Added get path for obtaining a model
- Moved model specific requests to  `download_model_get_request` and `search_model_get_request` to make former method a generic subpath e.g. recipes
- Added `download_model` and `download_model_from_stub` to directly download a model when returning the model isn't necessary. 
- Changed `load_model` and `load_model_from_stub` to use `get_model_get_request` for obtaining faster

### DIFF
--- a/src/sparsezoo/main.py
+++ b/src/sparsezoo/main.py
@@ -291,6 +291,11 @@ def parse_args():
         help="The directory to save the model files in, "
         "defaults to the cwd with the model description as a sub folder",
     )
+    download_parser.add_argument(
+        "--overwrite",
+        action="store_true",
+        help="Overwrites existing model files if previously downloaded",
+    )
 
     search_parser.add_argument(
         "--page", type=int, default=1, help="The page of search results to view"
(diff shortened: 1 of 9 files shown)


bfineran (dismissed):
LGTM. I like the decoupling of the existing download_get_request with model downloads.

src/sparsezoo/requests/download.py
```diff
@@ -25,32 +25,38 @@
...
     force_token_refresh: bool = False,
 ) -> Dict:
     """
-    Get a downloadable model from the sparsezoo for any objects matching the args
+    Get a downloadable object from the sparsezoo for any objects matching the args
 
+    :param base_url: the base url
     :param args: the model args describing what should be downloaded for
-    :param file_name: the name of the file, if any, to get download info for
+    :param sub_path: the sub path from the model path if any
```
markurtz, line 50: can we give a bit more of an explanation for when this would be used?

markurtz (changes requested):

src/sparsezoo/main.py
```diff
@@ -384,7 +389,7 @@ def main():
 
     if args.command == DOWNLOAD_COMMAND:
         LOGGER.info("Downloading files from model...")
-        model = Zoo.load_model(
+        dir_path = Zoo.download_model(
```
markurtz, line 392: Should update the variable and potentially the print statement since we're returning a model here again

markurtz (approved):

bfineran (approved):

PR #60, opened 2021-04-01
Recipe API Path integration

kevinaer:
- Integration of recipe related Zoo methods with the Recipe API Paths
- Zoo model method to obtain model from recipe
- Zoo model method to obtain base model from recipe
- `base_stub` field for recipes, pointing to a base model if it exists.

### DIFF
--- a/src/sparsezoo/models/zoo.py
+++ b/src/sparsezoo/models/zoo.py
@@ -985,9 +985,7 @@ class Zoo:
         else:
             # search for base model, and return those weights as a starting checkpoint
             if not recipe.base_stub:
-                raise ValueError(
-                    f"Could not find base model for model {model.model_metadata}"
-                )
+                raise ValueError(f"Could not find base model for recipe {recipe.stub}")
 
             base_model = Zoo.load_base_model_from_recipe(
                 recipe,
--- a/src/sparsezoo/models/zoo.py
+++ b/src/sparsezoo/models/zoo.py
@@ -27,9 +27,13 @@ from sparsezoo.objects.optimization_recipe import (
 )
 from sparsezoo.requests import (
     ModelArgs,
+    RecipeArgs,
     download_model_get_request,
+    download_recipe_get_request,
     get_model_get_request,
+    get_recipe_get_request,
     search_model_get_request,
+    search_recipe_get_request,
 )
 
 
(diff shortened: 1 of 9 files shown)


src/sparsezoo/models/zoo.py
```diff
@@ -186,6 +190,68 @@ def load_model_from_stub(
...
+            dataset=metadata.dataset,
+            training_scheme=metadata.training_scheme,
+            optim_name=metadata.optim_name,
+            optim_category=metadata.optim_category,
+            optim_target=metadata.optim_target,
+            release_version=metadata.release_version,
+            override_folder_name=override_folder_name,
+            override_parent_path=override_parent_path,
+            force_token_refresh=force_token_refresh,
+        )
```
bfineran, line 228: `model_metadata` is a subclass of ModelArgs so we should be able to feed this directly into `load_model_from_stub`

src/sparsezoo/models/zoo.py
```diff
@@ -670,17 +996,18 @@ def download_recipe_base_framework_files(
...
-                raise ValueError(f"Could not find base model for model {model}")
-            framework_files = base_model[0].download_framework_files(
-                extensions=extensions
+            base_model = Zoo.load_base_model_from_recipe(
+                recipe,
+                override_folder_name=override_folder_name,
+                override_parent_path=override_parent_path,
             )
 
+            if not recipe.base_stub:
```
bfineran, line 1005: should this go before the `load` call?

bfineran (commented):
looks really good. two small comments

src/sparsezoo/objects/file.py
```diff
@@ -75,11 +76,12 @@ class File(BaseObject, Downloadable):
 
     def __init__(
         self,
-        model_metadata: ModelMetadata,
+        model_metadata: Union[Dict, ModelMetadata],
         file_id: str,
         display_name: str,
         file_type: str,
         operator_version: Union[str, None],
+        release_version: Union[Dict, ReleaseVersion],
```
markurtz, line 84: need to add docs for release_version in class constructor

src/sparsezoo/objects/file.py
```diff
@@ -90,6 +92,12 @@ def __init__(
         override_parent_path: Union[str, None] = None,
         **kwargs,
     ):
+        if isinstance(model_metadata, dict):
```
markurtz, line 95: should likely have an assert here to fail for the case where release_version is passed in and does not override the model_metadata since that information will be dropped

src/sparsezoo/objects/optimization_recipe.py
```diff
@@ -53,6 +54,7 @@ def __init__(
         recipe_id: str,
         recipe_type: str,
         display_description: str,
+        base_stub: Optional[str],
```
markurtz, line 57: need to add docs for base_stub for class constructor

src/sparsezoo/objects/tag.py
```diff
@@ -24,13 +24,13 @@
...
     """
-    A tag for a Model or OptimizationRecipe
+    A tag for a Model or Recipe
 
     :param tag_id: the tag id
     :param display_name: the display name for tag
     :param model_id: the model id if the tag is associated with a Model
     :param recipe_id: the optimization id if tag is associated
-        with an OptimizationRecipe
+        with an Recipe
```
markurtz, line 33: NIT: a Recipe

src/sparsezoo/objects/tag.py
```diff
@@ -81,6 +81,6 @@ def model_id(self) -> str:
     @property
     def recipe_id(self) -> str:
         """
-        :return: the optimization id if tag is associated with an OptimizationRecipe
+        :return: the optimization id if tag is associated with an Recipe
```
markurtz, line 84: NIT: a Recipe

markurtz (changes requested):
Left a fee comments.

Note for the future, the Zoo class is quickly becoming populated with a lot of functions as well as functions with many params. In the near future I think it would be good if we start migrating support to being able to construct Model and Recipe objects from the metadata and have those run the get, download, etc to populate the rest of the info

src/sparsezoo/objects/file.py
kevinaer: The release version should only override when model_metadata is a dict that does not have release_version. Updated logic to make that more clear.

markurtz (approved):

bfineran (approved):

PR #64, opened 2021-04-06
Convenience methods for Models and Recipes

kevinaer:
- Added convenience methods for Model class to obtain similar/optimized models
- Added convenience method for Recipe to obtain Model
- Changed Zoo to use a factory approach for creating Recipe, Model, and RecipeTypes to circumvent circular import

Note: Waiting on [PR60](https://github.com/neuralmagic/sparsezoo/pull/60) to land

### DIFF
--- a/src/sparsezoo/models/zoo.py
+++ b/src/sparsezoo/models/zoo.py
@@ -93,20 +93,30 @@ class Zoo:
 
     @staticmethod
     def constructor(construct_key: str):
+        """
+        :param construct_key: The constructor key
+        :return: Constructor registered with the key
+        """
         if construct_key not in Zoo._CONSTRUCTORS:
             raise RuntimeError(f"No constructure registered for {construct_key}")
         return Zoo._CONSTRUCTORS[construct_key]
 
     @staticmethod
-    def _register_class(key: str, const_func: Callable):
+    def _register_constructor(key: str, const_func: Callable):
         Zoo._CONSTRUCTORS[key] = const_func
 
     @staticmethod
     def register(
         key: str,
     ):
+        """
+        Registers a constructor with key
+
+        :param key: The key the constructor will be registered with
+        """
+
         def decorator(const_func: Callable):
-            Zoo._register_class(key, const_func)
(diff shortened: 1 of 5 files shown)


bfineran (dismissed):

src/sparsezoo/models/zoo.py
```diff
@@ -89,6 +89,37 @@ class Zoo:
...
+    def constructor(construct_key: str):
+        """
+        :param construct_key: The constructor key
+        :return: Constructor registered with the key
+        """
+        if construct_key not in Zoo._CONSTRUCTORS:
+            raise RuntimeError(f"No constructure registered for {construct_key}")
+        return Zoo._CONSTRUCTORS[construct_key]
+
+    @staticmethod
```
markurtz, line 104: NIT: move private static method to after public static

markurtz (changes requested):
I think this might end up being a lot cleaner if we move the functionality of creating and searching for models and recipes into the respective classes on construction of them as well as static and instance methods on them for searching. Then we can have the zoo just call into the Model and Recipe classes rather than us going back and forth between these classes which can quickly become confusing an hard to debug.

On the topic of debugging, would be good to start adding in debug logs when constructing and calling any main functions

src/sparsezoo/models/zoo.py
```diff
@@ -17,77 +17,53 @@
...
-    if valid_params is not None and any(param not in valid_params for param in params):
-        warnings.warn(
-            f"Invalid query string for stub {stub} valid params include {valid_params},"
-            f" given {list(params.keys())}"
-        )
+    _CONSTRUCTORS = {}
 
-    return stub, params
+    @staticmethod
+    def constructor(construct_key: str):
```
markurtz, line 40: @kevinaer do we still need the constructor and register functions now?

kevinaer: Ah good catch, we do not. Will get rid of them

markurtz (approved):

bfineran (approved):

PR #69, opened 2021-04-15
SparseZoo Table UI

kevinaer:
Added a simple dynamic table for sparsezoo models. 

Core Features:
- View models by domain and subdomain. 
- Pagination of models
- Filtering based on fields

Remaining Work:
- Review docstring for functions
- GitHub action to deploy changes to S3 bucket (may end up in separate PR)
- Look into saving auth token in a cookie to reduce loading time (may end up in separate PR)

UI not finalized, comments made on UI should be left in this PR. They will be address on subsequent PR.

Including temporary urls for [image classifications](http://sparsezoo-table.neuralmagic.com.s3-website-us-east-1.amazonaws.com/models/cv/classification) and [object detection](http://sparsezoo-table.neuralmagic.com.s3-website-us-east-1.amazonaws.com/models/cv/detection)


src/ui/app/app-theme.jsx
```diff
@@ -0,0 +1,52 @@
...
+  const theme = createMuiTheme({
+    palette: {
+      type: darkMode ? "dark" : "light",
+      primary: { main: "#4652B1", contrastText: "#FFFFFF" },
+      secondary: { main: "#ff9900", contrastText: "#FFFFFF" },
+      disabled: { main: "#777777" },
+      divider: "#000000",
+    },
+    typography: {
+      fontFamily: ["Roboto", "Open Sans", "sans-serif"].join(","),
```
markurtz, line 34: Typography should default to Helveltica Neue first

src/ui/app/app.jsx
```diff
@@ -0,0 +1,77 @@
...
+        dispatch(setAuthToken(authToken));
+      } else {
+        dispatch(authThunk());
+      }
+    } else if (authState.status === "succeeded") {
+      const cookies = new Cookies();
+
+      const authToken = cookies.get("NM_AUTH_TOKEN");
+      if (authToken !== authState.token) {
+        const expires = moment().add(7, "days").toDate();
```
markurtz, line 53: can we get a lookup from the API for when the token is supposed to expire? This could cause us some issues in the future if we ever decide to change it

src/ui/components/filter-sidebar/filter-sidebar-styles.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+export default function makeModelTableStyles() {
+  return makeStyles(
+    (theme) => ({
+      root: {
+        height: "100%",
+        paddingLeft: theme.spacing(1),
+        overflowY: "scroll",
+        display: "inline-block",
+        backgroundColor: "rgba(0, 0, 0, 0.04)",
```
markurtz, line 27: can we get a setting in theme so this populates from there?

src/ui/components/filter-sidebar/filter-sidebar.jsx
```diff
@@ -0,0 +1,86 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from "react";
+
+import _ from "lodash";
```
markurtz, line 19: can we get an alias set for this? I'm not a fan of using single or even double character names because it makes searching on them much harder since it will match a lot of other text

src/ui/components/loader-overlay/loader-overlay-styles.jsx
```diff
@@ -0,0 +1,61 @@
...
+          backgroundColor: theme.palette.overlay,
+          color: theme.palette.primary.light,
+          zIndex: 1200,
+        },
+        loader: {
+          width: "100%",
+          height: "100%",
+        },
+        error: {
+          width: "80%!important",
```
markurtz, line 38: do we have to use important here or is there another way to get by without forcing it?

src/ui/components/model-table/model-table-styles.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { makeStyles } from "@material-ui/core/styles";
+
+export default function makeModelTableStyles() {
+  return makeStyles(
+    (theme) => ({
+      root: {
+        padding: theme.spacing(0.5),
+      },
+      gridItem: {
+        height: "485px",
```
markurtz, line 26: this seems really tall for an item

src/ui/components/zoo-table/zoo-table-headers/zoo-table-headers-styles.jsx
```diff
@@ -0,0 +1,47 @@
...
+export default function makeZooTableHeaderStyles() {
+  return makeStyles(
+    (theme) => ({
+      root: {},
+      header: {
+        fontWeight: "bold",
+        display: "inline",
+      },
+      headerContainer: {
+        backgroundColor: "white",
```
markurtz, line 28: can we get this to pull from the theme?

src/ui/components/zoo-table/zoo-table-styles.jsx
```diff
@@ -0,0 +1,60 @@
...
+      loaderText: {
+        color: theme.palette.info.contrastText,
+      },
+      pagination: {
+        borderStyle: "solid",
+        borderWidth: "0.5px 0 0 0",
+        borderColor: theme.palette.divider,
+      },
+      paginatedTableContainer: {
+        maxHeight: 37 + 330,
```
markurtz, line 32: why the addition here? Also, seems like an odd height

src/ui/components/zoo-table/zoo-table-styles.jsx
```diff
@@ -0,0 +1,60 @@
...
+      },
+      tableGroup: {
+        position: "relative",
+        borderStyle: "solid",
+        borderWidth: "thin",
+        border: theme.palette.divider,
+        borderBottomWidth: 0,
+      },
+      paginatedTableGroup: {
+        maxHeight: 37 + 330 + 52,
```
markurtz, line 46: why the addition and seems like an odd height

markurtz (changes requested):
A few minor things but overall looks good. We should definitely start thinking through what test cases would look like for these to keep it in good shape. Also, I think anything exported in these modules we should add docs for

src/ui/components/loader-overlay/loader-overlay-styles.jsx
kevinaer: This component was copied over from sparsify. Most likely not necessary

kevinaer:
At least from my previous experiences, enzyme with jest would be the best way to write tests for components. Definitely for post release will start filling out backlog of tests.

markurtz (approved):

bfineran (approved):

PR #101, opened 2021-07-14
Add:Support for batched iteration

rahul-tuli:
This pull request adds support for batched iteration,
Let me know if more tests are needed.

Implementation adapted from: [YoloV3 benchmarking DeepSparse](https://github.com/neuralmagic/deepsparse/blob/main/examples/ultralytics-yolov3/benchmark.py#L384)

### DIFF
--- a/src/sparsezoo/utils/data.py
+++ b/src/sparsezoo/utils/data.py
@@ -19,7 +19,7 @@ Utilities for data loading into numpy for use in ONNX supported systems
 import logging
 import math
 from collections import OrderedDict
-from typing import Dict, Iterable, Iterator, List, Tuple, Union
+from typing import Dict, Generator, Iterable, Iterator, List, Tuple, Union
 
 import numpy
 
@@ -27,7 +27,6 @@ from sparsezoo.utils.numpy import NumpyArrayBatcher, load_numpy_list
 
 
 __all__ = ["Dataset", "RandomDataset", "DataLoader"]
-
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tests/test_dataset.py
```diff
@@ -0,0 +1,72 @@
...
+    data_loader = _data.iter_batches(batch_size=batch_size, iterations=iterations)
+    data_shape = _data.data.shape
+
+    # fix 1-d numpy array shape
+    if len(data_shape) == 1:
+        data_shape = (data_shape[0], 1)
+
+    for iteration, batch in enumerate(data_loader):
+        batch_shape = batch[0].shape
+
```
bfineran, line 66: should add an assertion that the batch has the correct `batch_size`

tests/test_dataset.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: move file to `tests/sparsezoo/utils/test_data.py` to match file naming conventions

tests/test_dataset.py
```diff
@@ -0,0 +1,72 @@
...
+    ],
+)
+def test_has_iter_batches(_data):
+    assert hasattr(_data, "iter_batches")
+
+
+@pytest.mark.parametrize(
+    "_data",
+    [
+        Dataset(data=np.empty(shape=(100, 100, 10, 10)), name="4d"),
```
bfineran, line 34: input to dataset should be an iterable of numpy list(s), instead of a single one

bfineran (changes requested):

src/sparsezoo/utils/data.py
```diff
@@ -62,6 +61,49 @@ def __iter__(self) -> Iterator[Union[numpy.ndarray, Dict[str, numpy.ndarray]]]:
...
+        batch_size: int,
+        iterations: int,
+    ) -> Generator[List[numpy.ndarray], None, None]:
+        """
+        A method to iterate over data in batches
+
+        :param batch_size: Positive integer representing the size of each batch
+        :param iterations: Positive integer representing number of batches to yield
+
+        :pre: batch_size should be within the range [0, number_of_samples)
```
bfineran, line 75: `batch_size` can be any non-negative value since this implementation will keep looping over the dataset even if it starts repeating. this is good for cases where we have say one sample but want to measure across multiple batch sizes. no code change needed

src/sparsezoo/utils/data.py
```diff
@@ -62,6 +61,49 @@ def __iter__(self) -> Iterator[Union[numpy.ndarray, Dict[str, numpy.ndarray]]]:
...
+        :pre: iterations should be a positive integer
+
+        :return: A batch generator for self.data
+        """
+        dataset = self._data
```
bfineran, line 80: nit: refer to `self._data` directly since this points directly to it

tests/test_data.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: file should be moved to `tests/sparsezoo/utils/test_data.py`

bfineran (changes requested):

PR #102, opened 2021-07-16
Add support for batched Iteration 

rahul-tuli:
This pull request adds support for batched iteration in sparsezoo's Dataset class


src/sparsezoo/utils/data.py
```diff
@@ -28,10 +28,158 @@
...
+            self._batch_buffer.append(sample)
+            if self._buffer_is_full:
+                _batch = self._make_batch()
+                yield _batch
+                self._empty_buffer()
+                self._batches_created += 1
+                if self._all_batches_loaded:
+                    break
+
+    def _empty_buffer(self):
```
bfineran, line 143: nit: this is only used once, can we remove it and just write `self._batch_buffer = []` above?

src/sparsezoo/utils/data.py
```diff
@@ -76,6 +224,16 @@ def data(self) -> List[Union[numpy.ndarray, Dict[str, numpy.ndarray]]]:
         """
         return self._data
 
+    def iter_batches(
+        self, batch_size: int, iterations: int
+    ) -> Generator[List[numpy.ndarray], None, None]:
+        """
```
bfineran, line 230: nit: add params and return to docstring

tests/sparsezoo/utils/utilities.py
```diff
@@ -23,6 +23,14 @@
 from sparsezoo.utils import CACHE_DIR
```
bfineran, line 23: let's not include these changes right now, since they're not in the scope for testing `iter_batches`

tests/sparsezoo/utils/__init__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: let's not include these changes right now, since they're not in the scope for testing `iter_batches`

bfineran (commented):
nice refactor @rahul-tuli 
going to sync offline about the single input case

src/sparsezoo/utils/data.py
```diff
@@ -28,10 +28,158 @@
...
+        batch_size: int,
+        iterations: int,
+    ):
+        self._data = data
+        self._batch_size = batch_size
+        self._iterations = iterations
+        if batch_size < 0 or iterations < 0:
+            raise ValueError(
+                f"Both batch size and number of _iterations should be non-negative, "
+                f"supplied values (_batch_size, _iterations):{(batch_size, iterations)}"
```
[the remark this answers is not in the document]
rahul-tuli: Resolved!

src/sparsezoo/utils/data.py
rahul-tuli: Resolved

tests/sparsezoo/utils/utilities.py
rahul-tuli: Resolved

src/sparsezoo/utils/data.py
rahul-tuli: resolved

tests/sparsezoo/utils/__init__.py
rahul-tuli: Will have to move the tests/utils.py file, if we need to keep test_data.py in tests/utils/test_data.py
Can keep it as is if we can change the tests/utils folder to something else.

bfineran: good catch.  Can we rename `tests/utils.py` to `tests/helpers.py`? Will need to make sure to rename all instances of importing `test.utils` from tests to `test.helpers` (and rerun all sparsezoo tests to be safe)

src/sparsezoo/utils/data.py
```diff
@@ -28,10 +28,98 @@
...
+        while not self._all_batches_loaded:
+            yield from self._batch_generator(source=self._data)
+
+    def _batch_generator(self, source) -> Generator[List[numpy.ndarray], None, None]:
+        # batches from source
+        for sample in source:
+            self._batch_buffer.append(sample)
+            if self._buffer_is_full:
+                _batch = self._make_batch()
+                yield _batch
```
bfineran, line 94: nit: can we add in functionality to unwrap the list into a single value if originally there were single inputs? (i.e. if `self._data` is a `List[numpy.ndarray]` then unwrap the single list, but if it happens to be a `List[List[numpy.ndarray]]` with the inner list size 1, then don't unwrap. This can be checked in the initializer for `_BatchLoader`

bfineran (commented):
Looking good, just one comment for the code. @rahul-tuli for `tests/sparsezoo/utils.py` can we rename it to `helpers.py` and then we need to change every instance of `from tests.sparsezoo.utils import ...` to `from tests.sparsezoo.helpers import ...`

tests/sparsezoo/utils/__init__.py
rahul-tuli: Resolved!

src/sparsezoo/utils/data.py
rahul-tuli: Sure thang!

bfineran (approved):
LGTM pending tests!

bfineran (approved):

src/sparsezoo/utils/data.py
```diff
@@ -28,10 +28,103 @@
...
+    ]
+
+    def __init__(
+        self,
+        data: Iterable[Union[numpy.ndarray, List[numpy.ndarray]]],
+        batch_size: int,
+        iterations: int,
+    ):
+        self._data = data
+        self._single_input = type(self._data[0]) is numpy.ndarray
```
mgoin, line 55: this would still be true if someone passed in a `List[numpy.ndarray]`

src/sparsezoo/utils/data.py
```diff
@@ -28,10 +28,103 @@
...
+        if self._single_input:
+            self._data = [self._data]
+        self._batch_size = batch_size
+        self._iterations = iterations
+        if batch_size < 0 or iterations < 0:
+            raise ValueError(
+                f"Both batch size and number of iterations should be non-negative, "
+                f"supplied values (batch_size, iterations):{(batch_size, iterations)}"
+            )
```
mgoin, line 64: both of these should be greater than 0

src/sparsezoo/utils/data.py
```diff
@@ -28,10 +28,103 @@
...
+                self._batches_created += 1
+                if self._all_batches_loaded:
+                    break
+
+    def _init_batch_template(
+        self,
+    ) -> Iterable[Union[List[numpy.ndarray], numpy.ndarray]]:
+
+        # A placeholder for batches
+
```
mgoin, line 106: nit: remove empty newlines

src/sparsezoo/utils/data.py
```diff
@@ -28,10 +28,103 @@
...
+        # Copy contents of buffer to batch placeholder
+        # and return A list of numpy array(s) representing the batch
+
+        batch = [
+            numpy.stack([sample[idx] for sample in self._batch_buffer], out=template)
+            for idx, template in enumerate(self._batch_template)
+        ]
+
+        if self._single_input:
+            batch = batch[0]
```
mgoin, line 124: why does starting with a single input produce a single batch? this is ignoring the batch_size parameter

mgoin (changes requested):

src/sparsezoo/utils/data.py
rahul-tuli: Resolved!

src/sparsezoo/utils/data.py
rahul-tuli: It produces more batches, we are unwrapping an extra outer list here.

src/sparsezoo/utils/data.py
rahul-tuli: We want it to be True in that case, maybe inverting the condition and renaming the variable will make more sense, check latest push.

caveat: this is to distinguish b/w [[np.ndarray]] and [np.ndarray]

src/sparsezoo/utils/data.py
rahul-tuli: 👍

src/sparsezoo/utils/data.py
rahul-tuli: we wrap inputs of the form List[np.array] in an extra outer list in the initializer, then unwrap it just before yielding a batch

bfineran (approved):

src/sparsezoo/utils/data.py
bfineran: yup, if the dataset is of numpy arrays, we want to return batches of numpy arrays. this unwraps the array batch from a list as part of other shared logic

kevinaer (approved):

PR #103, opened 2021-08-02
Refactor: Remove _BatchLoader, add tests for DataLoader

rahul-tuli:
Also moved validation of of dataset, batch_size, iter_steps to top of init

### DIFF
--- a/tests/sparsezoo/utils/test_data.py
+++ b/tests/sparsezoo/utils/test_data.py
@@ -14,17 +14,21 @@
 
 
 from collections import OrderedDict
+from typing import Tuple
 
 import pytest
 
-from sparsezoo.utils.data import DataLoader, RandomDataset
+from sparsezoo.utils.data import DataLoader, Dataset, RandomDataset
 
 
 # TESTS for DataLoader
 
 
 @pytest.fixture
-def dataset():
+def dataset() -> Dataset:
+    """
+    A Dataset containing 4 samples of 3 channel images
+    """
     name, typed_shapes, num_samples = "rd1", {"inp": ([3, 224, 224], None)}, 4
     data = RandomDataset(name=name, typed_shapes=typed_shapes, num_samples=num_samples)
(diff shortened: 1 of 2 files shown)


tests/sparsezoo/utils/test_data.py
```diff
@@ -11,204 +11,115 @@
...
+    data = RandomDataset(name=name, typed_shapes=typed_shapes, num_samples=num_samples)
+    yield data
+    del data
 
 
 @pytest.fixture
-def single_input_dataset():
-    data = [np.random.rand(3, 2), np.random.rand(3, 2), np.random.rand(3, 2)]
-    return Dataset(data=data, name="single-input-test-dataset")
+def dataset_b():
```
kevinaer, line 35: small nit, but can we be more descriptive about the fixture?

tests/sparsezoo/utils/test_data.py
```diff
@@ -11,204 +11,115 @@
...
+from sparsezoo.utils.data import DataLoader, RandomDataset
+
+
+# TESTS for DataLoader
 
 
 @pytest.fixture
-def dummy_dataset():
-    return Dataset(data=[np.random.rand(100, 10)], name="dummy-dataset")
+def dataset():
```
kevinaer, line 27: Can we add return types to the fixture method definitions?

kevinaer (changes requested):
Nit: can we get return types for the test fixtures?

tests/sparsezoo/utils/test_data.py
rahul-tuli: Yes sure thing!

tests/sparsezoo/utils/test_data.py
rahul-tuli: Done!

kevinaer (approved):

bfineran (approved):

PR #114, opened 2021-11-03
version-api-pypi integration

xkazuka:
Completed: 
- Added get request for the version-api lambda's with the package_name and the package_version and temp holder of package_integration
 - Warns the user/client if they are not on the latest version
- Run the api call in the background process using thread
- Environmental variable to turn on/off the api-call
- Make a call to staging-api.neuralmagic.com and save client info to staging DB

To-dos:
- figure out how to put in the package_integration info
- Add integration to other packages

### DIFF
--- a/src/sparsezoo/__init__.py
+++ b/src/sparsezoo/__init__.py
@@ -33,5 +33,5 @@ from sparsezoo.package import check_package_version as _check_package_version
 
 _check_package_version(
     package_name=__name__ if is_release else f"{__name__}-nightly",
-    package_version=version_base,
+    package_version=version,
 )
--- a/src/sparsezoo/__init__.py
+++ b/src/sparsezoo/__init__.py
@@ -15,7 +15,8 @@
 # limitations under the License.
 
 """
-Functionality for accessing models, recipes, and supporting files in the SparseZoo
+- Functionality for accessing models, recipes, and supporting files in the SparseZoo
+- Notify the user the last pypi package version
 """
 
 # flake8: noqa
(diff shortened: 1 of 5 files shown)


src/sparsezoo/__init__.py
```diff
@@ -25,3 +26,59 @@
...
+from .requests.base import LATEST_PACKAGE_VERSION_URL
+
+import logging
+import requests
+import threading
+import os
+
+_LOGGER = logging.getLogger(__name__)
+
+def _version_check(package_name, package_version, package_integration):
```
markurtz, line 39: @kazukageorge let's make this base function public as the other repositories will be expected to call it. As mentioned in the other comment, we'll need to nest this in a separate, appropriate package and then import the callable in here and rename as private so we don't export it from the root as a public function

src/sparsezoo/__init__.py
```diff
@@ -25,3 +26,59 @@
...
+                    f"'pip install {checkedPackage['packageName']}'"\
+                    "command"
+                )
+                _LOGGER.warning(
+                    "WARNING: "\
+                    "To turn off version check, set an environmental variable " \
+                    "NM_VERSION_CHECK=false"\
+                )
+    except:
+        pass
```
markurtz, line 67: we should be logging this exception here as a warning rather than silently ignoring it so we can surface any issues with the API or the contained code

src/sparsezoo/__init__.py
```diff
@@ -25,3 +26,59 @@
...
+                    "WARNING: "\
+                    f"You are using {checkedPackage['packageName']} "\
+                    f"version {checkedPackage['userVersion']} "\
+                    f"however version {checkedPackage['latestVersion']} "\
+                    "is available.\n"\
+                    "You should consider upgrading via executing the "\
+                    f"'pip install {checkedPackage['packageName']}'"\
+                    "command"
+                )
+                _LOGGER.warning(
```
markurtz, line 61: Let's combine this with the previous log statement so we're not taking up multiple lines in the log history and make this statement easier to filter/find when going back through logs

src/sparsezoo/__init__.py
```diff
@@ -25,3 +26,59 @@
...
+                    "To turn off version check, set an environmental variable " \
+                    "NM_VERSION_CHECK=false"\
+                )
+    except:
+        pass
+
+package_integration = "HuggingFace"  # temp holder
+
+# Run thread in the background
+if os.getenv("NM_VERSION_CHECK") == None or os.getenv("NM_VERSION_CHECK").lower().strip() != 'false':
```
markurtz, line 72: Let's wrap these checks and kicking off of the background thread into a proper function as well to live alongside the package check request

src/sparsezoo/__init__.py
```diff
@@ -25,3 +26,59 @@
...
+    threading.Thread(
+        target=_version_check,
+        kwargs={
+            "package_name": __name__,
+            "package_version": version,
+            "package_integration": package_integration,
+        },
+    ).start()
+else:
+    print('skipping version check')
```
markurtz, line 82: this should always be in a logger rather than a print statement

markurtz (changes requested):

src/sparsezoo/package.py
```diff
@@ -0,0 +1,112 @@
...
+        for checked_package in response_json["checked_packages"]:
+            if not checked_package["is_latest"]:
+                LOGGER.warning(
+                    "WARNING: "
+                    f"You are using {checked_package['package_name']} "
+                    f"version {checked_package['user_package_version']} "
+                    f"however version {checked_package['latest_package_version']} "
+                    "is available.\n"
+                    "Consider upgrading via executing the "
+                    f"'pip install {checked_package['package_name']}' "
```
bfineran, line 58: should be `pip install --upgrade`

bfineran (dismissed):

markurtz (approved):

src/sparsezoo/__init__.py
```diff
@@ -25,3 +26,11 @@
 from .main import *
 from .models.zoo import *
 from .objects import *
+from .package import *
+
+
+from sparsezoo.package import check_package_version as _check_package_version
+
+_check_package_version(
+    package_name=__name__, package_integration="Null", package_version=version_base
```
markurtz, line 35:
```suggestion
    package_name=__name__, package_integration=None, package_version=version_base
```

markurtz (changes requested):
small change on None and currently we are not going to be pushing up the nightly tag

markurtz (dismissed):

markurtz (approved):

PR #119, opened 2022-01-05
Mark recently intermittent tests as flaky

dbarbuzzi:
In order to prevent any transient network issues from showing false failures in our standard test runs, the tests that recently intermittently failed have been decorated as `@flaky(max_runs=3)`.

ℹ️ **Note:** This required adding `flaky` as a new dev dependency to the repo.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ ensure_newline_before_comments = True
 force_grid_wrap = 0
 include_trailing_comma = True
 known_first_party = sparsezoo,tests
-known_third_party = bs4,requests,packaging,yaml,tqdm,numpy,onnx,pytest
+known_third_party = bs4,requests,packaging,yaml,tqdm,numpy,onnx,pytest,flaky
 sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
 
 line_length = 88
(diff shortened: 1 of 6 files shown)


rahul-tuli (approved):

bfineran (approved):
LGTM, but wondering any reason 3 is chosen? If we are only looking to ignore transient failures, shouldn't 1/2 be enough?

dbarbuzzi:
> @bfineran 
> LGTM, but wondering any reason 3 is chosen? If we are only looking to ignore transient failures, shouldn't 1/2 be enough?

@bfineran Good question. This was the suggested amount during the discussion with @markurtz @kazukageorge and @dhuangnm. Mark or others, please correct me if misremembered.

dhuangnm:
> > @bfineran
> > LGTM, but wondering any reason 3 is chosen? If we are only looking to ignore transient failures, shouldn't 1/2 be enough?
> 
> @bfineran Good question. This was the suggested amount during the discussion with @markurtz @kazukageorge and @dhuangnm. Mark or others, please correct me if misremembered.

I guess we can set it to 2 for now since it's likely to pass the second time.

bfineran (approved):

rahul-tuli (approved):

PR #132, opened 2022-02-24
checkpoint download api

xkazuka:
Blocked on https://github.com/neuralmagic/zoomodels/pull/146

Added:
- [X] Show all checkpoints folders 
- [X] Download checkpoints .tar.gz folders with stub 

```python3
from sparsezoo.models import Zoo

stub = "zoo:test_nlp/masked_language_modeling/bert-base/pytorch/huggingface/wikipedia_bookcorpus/base-none"

print(Zoo.get_checkpoints_folders_from_stub(stub)) # show all checkpoints folders
Zoo.download_checkpoints_folders_from_stub(stub) # download all checkpoint folders
Zoo.download_checkpoints_folders_from_stub(
    stub=stub, 
    download_folders=[
            "checkpoints_epoch10.tar.gz",
            "checkpoints_epoch20.tar.gz",
        ]
    ) # download only the specific folders if they exist
```

### DIFF
--- a/src/sparsezoo/objects/model.py
+++ b/src/sparsezoo/objects/model.py
@@ -548,14 +548,12 @@ class Model(Downloadable, ModelMetadata):
         model = Model(
             **response_json["model"],
         )
-        return (
-            [
-                checkpoints_file["display_name"]
-                for checkpoints_file in model.checkpoints_folders
-            ]
-            if model.checkpoints_folders
-            else None
-        )
+        ckpt_files = []
+        if model.checkpoints_folders:
+            for checkpoints_file in model.checkpoints_folders:
+                ckpt_file_metadata = checkpoints_file.dict()
+                ckpt_files.append(ckpt_file_metadata["display_name"])
+        return ckpt_files
 
     @staticmethod
     def search_models(
--- a/src/sparsezoo/objects/model.py
+++ b/src/sparsezoo/objects/model.py
@@ -511,7 +511,10 @@ class Model(Downloadable, ModelMetadata):
 
         if isinstance(stub, str):
             stub, _ = parse_zoo_stub(stub, valid_params=[])
-        _LOGGER.debug(f"download_model_from_stub: downloading model from stub {stub}")
+        _LOGGER.debug(
+            "download_checkpoints_folders_from_stub: "
+            f"searching for model withs stub: {stub}"
+        )
         response_json = download_model_get_request(
             args=stub,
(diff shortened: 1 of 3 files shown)


src/sparsezoo/objects/model.py
```diff
@@ -164,6 +164,21 @@ def __init__(
             for res in sorted_files[FileTypes.FRAMEWORK.value]
         ]
 
+        self._checkpoints_folders = (
+            [
+                File(
+                    model_metadata=metadata,
+                    # child_folder_name=metadata.checkpoints,
```
markurtz, line 171: remove commented code

src/sparsezoo/objects/model.py
```diff
@@ -887,6 +972,73 @@ def download_framework_files(
...
+                        overwrite=overwrite,
+                        refresh_token=refresh_token,
+                        show_progress=show_progress,
+                    )
+                    downloaded_paths.append(folder.path)
+            return downloaded_paths
+
+        return None
+
+        # for file in self._framework_files:
```
markurtz, line 1024: Remove commented code

src/sparsezoo/objects/model.py
```diff
@@ -164,6 +164,21 @@ def __init__(
...
+                    model_metadata=metadata,
+                    # child_folder_name=metadata.checkpoints,
+                    override_folder_name=override_folder_name,
+                    override_parent_path=override_parent_path,
+                    **res,
+                )
+                for res in sorted_files[FileTypes.CHECKPOINTS.value]
+            ]
+            if sorted_files[FileTypes.CHECKPOINTS.value]
+            else None
```
markurtz, line 179: let's fall back on an empty array rather than None so reference code can loop through safely without a truth check on it

src/sparsezoo/objects/model.py
```diff
@@ -887,6 +972,73 @@ def download_framework_files(
             downloaded_paths.append(file.path)
         return downloaded_paths
 
+    def download_checkpoints_folders(
```
markurtz, line 975: let's clean up this function a bit -- it's very nested and very long meaning there are a lot of complicated flows in this that are hard to read through

markurtz (changes requested):
@kazukageorge for the use cases, the most important is allowing us to specify a specific checkpoint to download rather than downloading all of the checkpoints and knowing whether that was downloaded or not.

Additionally left some comments

PR #147, opened 2022-05-11
[Research CI/CD] Factory methods for ModelDirectory

dbogunowicz:


src/sparsezoo/refactor/directory.py
```diff
@@ -0,0 +1,194 @@
...
+                dest_path=os.path.join(destination_path, self.name),
+                overwrite=overwrite,
+            )
+            self.path = os.path.join(destination_path, self.name)
+
+        # Directory can represent a folder.
+        else:
+            for file in self.files:
+                if isinstance(file, Directory):
+                    raise NotImplementedError(
```
bfineran, line 104: any reason we can't recurse?

src/sparsezoo/refactor/directory.py
```diff
@@ -0,0 +1,194 @@
...
+# limitations under the License.
+
+import logging
+import os
+import pathlib
+import tarfile
+from typing import List, Optional
+
+from src.sparsezoo.refactor.file import File
+from src.sparsezoo.utils.downloader import download_file
```
bfineran, line 22: need to import from sparsezoo, not src.sparsezoo

bfineran, line 22: need to import from sparsezoo, not src.sparsezoo

src/sparsezoo/refactor/directory.py
```diff
@@ -0,0 +1,194 @@
...
+        else:
+            for file in self.files:
+                if isinstance(file, Directory):
+                    raise NotImplementedError(
+                        "Method download() does not support nested directories yet."
+                    )
+                else:
+                    file.download(destination_path=destination_path)
+
+    @classmethod
```
bfineran, line 110: why make this a classmethod if we are taking a `Directory` as input? any reason to not do `self.gzip`?

bfineran, line 110: why make this a classmethod if we are taking a `Directory` as input? any reason to not do `self.gzip`?

src/sparsezoo/refactor/directory.py
```diff
@@ -0,0 +1,194 @@
...
+
+        tar_file_name = directory.name + ".tar.gz"
+        tar_file_path = os.path.join(parent_path, tar_file_name)
+        with tarfile.open(tar_file_path, "w") as tar:
+            for file in directory.files:
+                tar.add(file.path)
+        return Directory(name=tar_file_name, path=tar_file_path)
+
+    @classmethod
+    def unzip(
```
bfineran, line 150: why make this a classmethod if we are taking a `Directory` as input? any reason to not do `self.unzip`?

bfineran, line 150: why make this a classmethod if we are taking a `Directory` as input? any reason to not do `self.unzip`?

src/sparsezoo/refactor/directory.py
```diff
@@ -0,0 +1,194 @@
...
+            )
+        tar.close()
+
+        return Directory(name=name, files=files, path=path)
+
+    def __len__(self):
+        return len(self.files)
+
+    def _is_tar(self):
+        _, *extension = self.name.split(".")
```
bfineran, line 193: having multiple `.`s outside of extension is valid in file and directory names, for example, we could have `my-model-version-1.5.3-pruned.tar.gz`.  would suggest `extension[-2:]`

bfineran, line 193: having multiple `.`s outside of extension is valid in file and directory names, for example, we could have `my-model-version-1.5.3-pruned.tar.gz`.  would suggest `extension[-2:]`

src/sparsezoo/refactor/model_directory.py
```diff
@@ -0,0 +1,516 @@
...
+
+FRAMEWORKS = ["pytorch", "keras", "tensorflow"]
+ENGINES = ["onnxruntime", "deepsparse"]
+
+
+def file_dictionary(**kwargs):
+    return kwargs
+
+
+class ModelDirectory:
```
bfineran, line 44: is there a reason this doesn't inherit from `Directory`?

bfineran, line 44: is there a reason this doesn't inherit from `Directory`?

src/sparsezoo/refactor/file.py
```diff
@@ -0,0 +1,208 @@
...
+import logging
+import os
+from typing import Any, Dict, Optional
+
+import onnx
+import yaml
+
+from PIL import Image
+from src.sparsezoo.utils.downloader import download_file
+from src.sparsezoo.utils.numpy import load_numpy_list
```
bfineran, line 25: need to import from sparsezoo, not src.sparsezoo

bfineran, line 25: need to import from sparsezoo, not src.sparsezoo

src/sparsezoo/refactor/model_directory.py
```diff
@@ -0,0 +1,516 @@
...
+from collections import OrderedDict
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import numpy as np
+import onnx
+
+import onnxruntime as ort
+from src.sparsezoo.refactor.directory import Directory
+from src.sparsezoo.refactor.file import File
+from src.sparsezoo.refactor.model_objects import (
```
bfineran, line 29: need to import from sparsezoo, not src.sparsezoo

bfineran, line 29: need to import from sparsezoo, not src.sparsezoo

src/sparsezoo/refactor/model_directory.py
```diff
@@ -0,0 +1,516 @@
...
+        self.model_card: File = self._file_from_files(
+            files, display_name="model.md"
+        )  # model.md
+        self.recipes: List[File] = self._file_from_files(
+            files, display_name="recipe(.*).md", regex=True
+        )  # recipe{_tag}.md
+
+        self._directory_path = directory_path
+
+    def __iter__(self) -> Tuple[str, Union[File, List[File], Dict]]:
```
bfineran, line 131: what would be the use case for this iter func?

bfineran, line 131: what would be the use case for this iter func?

bfineran (changes requested):
half way through review, will review the rest tomorrow

src/sparsezoo/refactor/model_directory.py
dbogunowicz: Clearly, it is desirable to have the opportunity to iterate over all the files in the ModelDirectory:
- "weak" reason - when running tests, I use this iterator to sequentially assert that all the attributes are properly setup.
- "strong" reason  - I use this `__iter__()` method in `ModelDirectory.validate()` and `ModelDirectory.download()` to iterate over all attributes to sequentially validate/attempt to download them.

Not sure if this is the best way of realizing this goal, open to suggestions.

dbogunowicz: See the answer below.

src/sparsezoo/refactor/directory.py
dbogunowicz: We can and we should. Overlooked this, fix in progress.

src/sparsezoo/refactor/directory.py
dbogunowicz: Fixed.

dbogunowicz: Fixed.

src/sparsezoo/refactor/directory.py
dbogunowicz: Fixed.

dbogunowicz: Fixed.

src/sparsezoo/refactor/directory.py
dbogunowicz: Good catch.

dbogunowicz:
@bfineran addressed all your comments(including moving files to a new location and setting a shared branch for all the incoming PRs).

bfineran:
@dbogunowicz looks like there is a test failing since the core API now depends on more external packages (here `PIL`) Let's figure out dependencies in anoter PR to `feature/ci-cd-refactor/main`

src/sparsezoo/v2/directory.py
```diff
@@ -0,0 +1,215 @@
...
+    def __init__(
+        self,
+        name: str,
+        files: Optional[List[File]] = None,
+        path: Optional[str] = None,
+        url: Optional[str] = None,
+        # TODO: Adding unpack as a parameter, mainly because of testing
+        # limitations (e.g. I cannot unpack tar archives which I try to
+        # download from SparseZoo but they are not there yet).
+        # We may remove this argument later.
```
bfineran, line 49: can we detect this and set upack accordingly?

src/sparsezoo/v2/directory.py
```diff
@@ -0,0 +1,215 @@
...
+        if self._is_archive and unpack:
+            self.unzip()
+
+    @property
+    def is_archive(self) -> bool:
+        """
+        Boolean flag:
+        - True if the Directory is an archive (tar)
+        - False if the Directory is a local directory (folder).
+        """
```
bfineran, line 68: nit: docstring needs return:

src/sparsezoo/v2/directory.py
```diff
@@ -0,0 +1,215 @@
...
+        return self._is_archive
+
+    @is_archive.setter
+    def is_archive(self, value: bool):
+        self._is_archive = value
```
bfineran, line 73: nit: needs docstring

src/sparsezoo/v2/directory.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def get_file_names(self) -> List[str]:
+        """
+        Get the names of the files in the Directory.
+        return: List with names of files
```
bfineran, line 78:
```suggestion
        Get the names of the files in the Directory.

        :return: List with names of files
```

src/sparsezoo/v2/file.py
```diff
@@ -0,0 +1,208 @@
...
+        Factory method for creating a File class object
+        from a file dictionary
+        (useful when working with the `request_json` from NeuralMagic).
+        """
+
+        name = file.get("display_name")
+        path = file.get("path")
+        url = file.get("url")
+
+        return File(
```
bfineran, line 82:
```suggestion
        return cls(
```

bfineran (commented):
LGTM overall, but there are quite a few TODOs. Do we plan on merging with those in? 

I've opened a 'SparseZoo gap analysis' task in asana [internal], could you update that first with any TODOs or open issues in the implementation and then (after 0.13 cut) we will add in any features available in `sparsezoo.Zoo` that are not currently available in the refactor to that ticket as well

src/sparsezoo/v2/model_directory.py
```diff
@@ -0,0 +1,514 @@
...
+
+    - `from_directory()` -> by using the directory to the model on your machine.
+
+    :param files: list of files, where every file
+        is represented by a dictionary (note: not Fil object)
+    :directory_path: if ModelDirectory created using method `from_directory()`,
+        it points to the directory of the ModelDirectory. By default: None
+    """
+
+    ATTRIBUTES = [
```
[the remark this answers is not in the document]
dbogunowicz: This should have been removed together with the `__iter__()` method of the `ModelDirectory`. Thank's for spotting it.

src/sparsezoo/v2/file.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        Factory method for creating a File class object
+        from a file dictionary
+        (useful when working with the `request_json` from NeuralMagic).
+        """
+
+        name = file.get("display_name")
+        path = file.get("path")
+        url = file.get("url")
+
```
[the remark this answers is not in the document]
dbogunowicz: To me validation makes sense if (exists `name`) AND (exists `path` OR exists `file`). A file definitely needs a name and some pointer to its contents (either local path or URL). I will include this in the next commit.

src/sparsezoo/v2/directory.py
dbogunowicz: Good idea, this should be beneficial.

src/sparsezoo/v2/file.py
```diff
@@ -0,0 +1,217 @@
...
+                "the file contents from. However, `url` is None."
+            )
+
+        if self.path is not None:
+            logging.warning(
+                f"Overwriting the current location of the File: {self.path} "
+                f"with the new location: {new_file_path}."
+            )
+
+        download_file(
```
rahul-tuli, line 112: I have a suggestion,
`download_file` in it's current implementation doesn't seem robust enough, What if the download fails? or we get `403`, `504` kind of errors,

Would it make sense, to wrap either `download_file` or `download` function here in a `retry` decorator?

The idea being wherever network is involved we try a few times before throwing an error to the user. My idea is something like the following from [`tenacity`](https://tenacity.readthedocs.io/en/latest/), we could implement a smaller version of this decorator on our own:

 ```python
 @retry(stop=(stop_after_delay(10) | stop_after_attempt(5)))
def stop_after_10_s_or_5_retries():
    print("Stopping after 10 seconds or 5 retries")
    raise Exception
 ``` 
 
 Does this make sense here?

src/sparsezoo/v2/model_directory.py
```diff
@@ -0,0 +1,502 @@
...
+        """
+        files = []
+        paths = glob.glob(os.path.join(directory_path, "*"))
+        for path in paths:
+            display_name = os.path.basename(path)
+            files.append(file_dictionary(display_name=display_name, path=path))
+
+        return ModelDirectory(files=files, name="model_directory", path=directory_path)
+
+    def download(self, directory_path: str, override: bool = False) -> bool:
```
rahul-tuli, line 158: same retry functionality maybe?

src/sparsezoo/v2/model_directory.py
```diff
@@ -0,0 +1,502 @@
...
+
+        validations = {}
+        for attribute, file in self.__iter__():
+            # TODO: Continuing with the hack
+            if attribute in SKIP_ATTRIBUTES:
+                validations[attribute] = True
+            elif isinstance(file, File):
+                validations[attribute] = file.validate()
+            elif isinstance(file, list):
+                validations[attribute] = all([_file.validate() for _file in file])
```
rahul-tuli, line 218:
```suggestion
                validations[attribute] = all(_file.validate() for _file in file)
```

This will create a generator that validates one file at a time and returns False when the first file is invalid; using a list will force it's creation entailing that `_file.validate()` is called for every file even if a previous file was invalid.
Thoughts?

src/sparsezoo/v2/model_directory.py
```diff
@@ -0,0 +1,502 @@
...
+        Then, feeds the data (sample original inputs)
+        to generate model outputs (in the iterative fashion).
+        :params engine_type: name of the inference engine
+        :returns engine output.
+        """
+        if engine_type not in ENGINES:
+            raise ValueError(f"The argument `engine_type` must be one of {ENGINES}")
+        elif engine_type == "onnxruntime":
+            for output in self._run_with_onnx_runtime():
+                yield output
```
rahul-tuli, line 242: noice! I love the use of `yield` here

src/sparsezoo/v2/model_directory.py
```diff
@@ -0,0 +1,502 @@
...
+            if file.get("contents"):
+                files_within = file["contents"]
+
+            # is directory locally on the machine
+            else:
+                paths_within = glob.glob(os.path.join(path, "*"))
+                files_within = [
+                    file_dictionary(display_name=os.path.basename(path), path=path)
+                    for path in paths_within
+                ]
```
rahul-tuli, line 297: nit:
```suggestion
                files_within = (
                    file_dictionary(display_name=os.path.basename(path), path=path)
                    for path in paths_within
                )
```
Can be converted to a generator again to `yield` values only when needed

src/sparsezoo/v2/model_directory.py
```diff
@@ -0,0 +1,502 @@
...
+        if sample_outputs.is_archive:
+            sample_outputs.unzip()
+
+        validation = []  # validation boolean per output
+        for target_output, output in zip(sample_outputs, self._run_with_onnx_runtime()):
+            target_output = list(target_output.values())
+            # TODO: This does not work for now, the
+            #  outputs are quite different. To be investigated.
+            validation.append(
+                all([np.array_equal(x, y) for x, y in zip(target_output, output)])
```
rahul-tuli, line 406: nit: Same as above
```suggestion
                all(np.array_equal(x, y) for x, y in zip(target_output, output))
```

src/sparsezoo/v2/model_directory.py
```diff
@@ -0,0 +1,502 @@
...
+                file.download(destination_path=directory_path)
+                return True
+            else:
+                logging.warning(
+                    f"Failed to download file {file.name}. The url of the file is None."
+                )
+                return False
+
+        elif isinstance(file, list):
+            validations = [self._download(_file, directory_path) for _file in file]
```
rahul-tuli, line 493: nit:
```suggestion
            validations = (self._download(_file, directory_path) for _file in file)
```

src/sparsezoo/v2/model_objects.py
```diff
@@ -0,0 +1,250 @@
...
+                validations[file.name] = self._check_file(file=file)
+
+        if not all(validations.values()):
+            logging.warning(
+                f"Following files: "
+                f"{[key for key, value in validations.items() if not value]} "
+                "were not successfully validated."
+            )
+
+        return all(validations.values())
```
rahul-tuli, line 84: Super bad nit: Can be ignored

Re-computation of result, could we store `all(validation.values())` in a variable and return that

src/sparsezoo/v2/model_objects.py
```diff
@@ -0,0 +1,250 @@
...
+                if file.name == file_name:
+                    return file
+                else:
+                    file = file.get_file(file_name=file_name)
+                    if file:
+                        return file
+
+            elif file.name == file_name:
+                return file
+
```
rahul-tuli, line 104: nit: Could avoid `if .. elif` ladder with early return

```suggestion
        for file in self.files:
            if file.name == file_name:
                    return file
            if isinstance(file, Directory):
                    file = file.get_file(file_name=file_name)
                    if file:
                        return file
```

src/sparsezoo/v2/model_objects.py
```diff
@@ -0,0 +1,250 @@
...
+    def __init__(
+        self,
+        files: List[File],
+        name: str,
+        path: Optional[str] = None,
+        url: Optional[str] = None,
+    ):
+        super().__init__(files=files, name=name, path=path, url=url)
+
+    def __iter__(self) -> File:
```
rahul-tuli, line 164: nice!

src/sparsezoo/v2/model_objects.py
```diff
@@ -0,0 +1,250 @@
...
+                    "Could not validate an NumpyDirectory given the provided onnx model."  # noqa: E501
+                )
+                return False
+
+        return True
+
+    def __iter__(self) -> OrderedDict:
+        for file in self.files:
+            for numpy_dict in load_numpy_list(file.path):
+                yield numpy_dict
```
rahul-tuli, line 223: nice!

rahul-tuli (commented):
Left a few nits and suggestion. Great job with code!

src/sparsezoo/v2/model_objects.py
dbogunowicz: good catch

KSGulin (commented):
Couple of nits, but looks great otherwise. Some very well organized code

src/sparsezoo/v2/file.py
```diff
@@ -0,0 +1,217 @@
...
+
+class File:
+    """
+    Object to wrap around common files. Currently, supporting:
+    - numpy files
+    - onnx files
+    - markdown files
+    - json files
+    - csv files
+    - image files
```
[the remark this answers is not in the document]
dbogunowicz: I don't think this is quite correct. `Directory` inherits from `File`, it is a broader concept.

If you have a numpy file or an onnx model, those shall be represented by the `File` object. Basically any `File` which cannot be divided further.
However, tar archives and folders would be represented by a concept "one rung up" the ladder of abstraction - the `Directory`.

src/sparsezoo/v2/model_objects.py
dbogunowicz: Haha, this is absolutely true. However, given that `validation` is a `dict`, it will not gives us significant computational overhead. I will keep the current version of the code, as I think it gives better readability.

src/sparsezoo/v2/model_directory.py
dbogunowicz: Hmm, thank you for bringing this up. I read up on the topic (e.g. https://stackoverflow.com/questions/47789/generator-expressions-vs-list-comprehensions) and I guess we can derive a heuristic:

If your computation will be super memory-heavy -> make sense to swap list comprehension to generator expression.
If you want to reuse the results of your computation -> you need to do list comprehension.

All in all @rahul-tuli while I don't think that those changes allow us to gain significant speedup, I am very grateful for those comments, definitely, I learned something there. Definitely please throw some more of those gems in the future, I enjoy learning :)

src/sparsezoo/v2/file.py
```diff
@@ -0,0 +1,217 @@
...
+    def from_dict(cls, file: Dict[str, Any]) -> "File":
+        """
+        Factory method for creating a File class object
+        from a file dictionary
+        (useful when working with the `request_json` from NeuralMagic).
+        :param file: a dictionary which contains an information about
+            the file (as returned by NeuralMagic API)
+        :return: File class object
+        """
+        name = file.get("display_name")
```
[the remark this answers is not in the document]
dbogunowicz: Wouldn't this be a bit redundant? I mean this information is already propagated in the type hint of the method's argument right?

KSGulin (approved):

src/sparsezoo/v2/file.py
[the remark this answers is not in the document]
rahul-tuli: I agree but types aren't enforced in Python. I guess a `ValueError` would be nicer and more informative for a user rather than the `AttributeError` which would get thrown when python can't find `get` attribute on the `file` object

src/sparsezoo/v2/model_directory.py
rahul-tuli: This is correct, if you look closely here though, we are not re-using the computation, the actual list is never used again, just the result from `all(...)` but I agree this will not have a big enough impact here.

src/sparsezoo/v2/directory.py
```diff
@@ -0,0 +1,222 @@
...
+        if not all(validations.values()):
+            logging.warning(
+                "Following files: "
+                f"{[key for key, value in validations.items() if not value]} "
+                "were not successfully validated."
+            )
+
+        return all(validations.values())
+
+    @retry(retries=3, retry_sleep_sec=5)
```
rahul-tuli, line 113: Yay!

src/sparsezoo/v2/file.py
```diff
@@ -0,0 +1,256 @@
...
+__all__ = ["File"]
+
+
+def retry(retries: int, retry_sleep_sec: int):
+    """
+    Retry Decorator
+    Retries the wrapped function/method
+        - `retry_num` times if the exceptions listed
+        -  with `retry_sleep_sec` interval between each attempt
+    :param retries: The number of times to repeat the wrapped function/method
```
rahul-tuli, line 40: nits return:
```suggestion
    
    :param retries: The number of times to repeat the wrapped function/method
```

src/sparsezoo/v2/file.py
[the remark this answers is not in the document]
rahul-tuli: I disagree upto an extent, a `Directory` should also be a `File` cause that's how it is in the real world. But this is not a blocker from my end

src/sparsezoo/v2/file.py
```diff
@@ -28,6 +31,40 @@
...
+                    logging.error(err)
+                    logging.error(traceback.format_exc())
+                    time.sleep(retry_sleep_sec)
+                logging.error("Trying attempt %s of %s.", attempt + 1, retries)
+            logging.error("Func %s retry failed", func)
+            raise Exception("Exceed max retry num: {} failed".format(retries))
+
+        return wrapper
+
+    return decorator
```
rahul-tuli, line 65: neat!

src/sparsezoo/v2/model_directory.py
```diff
@@ -404,19 +407,15 @@ def _validate_with_onnx_runtime(self):
             # TODO: This does not work for now, the
             #  outputs are quite different. To be investigated.
             validation.append(
-                all([np.array_equal(x, y) for x, y in zip(target_output, output)])
+                all((np.array_equal(x, y) for x, y in zip(target_output, output)))
```
rahul-tuli, line 410: This will work, a small nit for future: You don't need generator braces `()` inside `all`; as in you could do something like:
 `all(np.array_equal(x, y) for x, y in zip(target_output, output))`

rahul-tuli (approved):
To the moon! 🚀  @dbogunowicz

src/sparsezoo/v2/model_directory.py
dbogunowicz: yummy

src/sparsezoo/v2/directory.py
bfineran, line 113: @rahul-tuli @dbogunowicz great idea here - my worry is making this default behavior that the user can't override.  Let's get this added in instead as a parameter to download (can use recursion or something similar), maybe default to 1 that way the user can control the amount of retries and not take a max of 15 seconds + time to fail before exiting by default

bfineran: you can probably keep the retry decorator, just have it read max retries in from the function parameters and default to 1

src/sparsezoo/v2/directory.py
```diff
@@ -0,0 +1,246 @@
...
+                    f"Trying attempt {attempt + 1} of {retries}.", attempt + 1, retries
+                )
+            logging.error("Download retry failed...")
+            raise Exception("Exceed max retry attempts: {} failed".format(retries))
+
+        # Directory can represent a folder.
+        else:
+            for file in self.files:
+                if isinstance(file, Directory):
+                    raise NotImplementedError(
```
bfineran, line 159: note to add in for later: can just recurse here

src/sparsezoo/v2/file.py
```diff
@@ -0,0 +1,240 @@
...
+        from a file dictionary
+        (useful when working with the `request_json` from NeuralMagic).
+
+        :param file: a dictionary which contains an information about
+            the file (as returned by NeuralMagic API)
+        :return: File class object
+        """
+        name = file.get("display_name")
+        path = file.get("path")
+        url = file.get("url")
```
bfineran, line 85:
```suggestion
        name = file["display_name"]
        path = file.get("path")
        url = file.get("url")
```

bfineran (approved):
Approving - noting that we are landing to feature branch and have synced offline and via asana on TODOs and follow ups. Great work here @dbogunowicz

PR #156, opened 2022-05-26
[CI-CD Refactor] Swap PIL dependency for matplotlib

dbogunowicz:
Substituting PIL functionality with matplotlib. It is a much more "fundamental" / generic library.

### DIFF
--- a/tests/sparsezoo/v2/test_file.py
+++ b/tests/sparsezoo/v2/test_file.py
@@ -23,6 +23,7 @@ import pytest
 import yaml
 from onnx.helper import make_graph, make_model, make_node, make_tensor_value_info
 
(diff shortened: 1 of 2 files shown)


tests/sparsezoo/v2/test_file.py
```diff
@@ -23,6 +23,7 @@
 import yaml
 from onnx.helper import make_graph, make_model, make_node, make_tensor_value_info
 
+import matplotlib.pyplot as plt
 from PIL import Image
```
bfineran, line 27: we should remove this right?

KSGulin (approved):

bfineran (approved):

PR #158, opened 2022-05-26
[CI-CD Refactor] Recursion on download method

dbogunowicz:
I am not able to strictly test it yet, but I am fairly confident that this simple change does solve the problem.

### DIFF
--- a/src/sparsezoo/v2/directory.py
+++ b/src/sparsezoo/v2/directory.py
@@ -1,0 +1,246 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


bfineran (approved):

dbogunowicz:
@bfineran I am tracking new TODOs in Asana, will share it with you on Wednesday during our ML Eng meeting.

src/sparsezoo/v2/directory.py
```diff
@@ -152,15 +152,10 @@ def download(
             logging.error("Download retry failed...")
             raise Exception("Exceed max retry attempts: {} failed".format(retries))
 
-        # Directory can represent a folder.
+        # Directory can represent a folder or directory.
         else:
             for file in self.files:
```
KSGulin, line 157: I believe this will fail for File objects, because they don't have self.files

KSGulin (commented):
I like the recursion approach. Just needs a logic update

PR #160, opened 2022-05-27
[CI-CD Refactor] Validation of model cards/recipes 

dbogunowicz:
Enable `File` class object to parse `.md` files (either model cards or recipes).
Note, that validation of recipes is optional. This is because to validate a recipe, we need to use tools present in SparseML, and we want to avoid explicit dependency on the library.


bfineran (commented):
lgtm pending comment

KSGulin (approved):

bfineran (approved):

PR #161, opened 2022-06-01
[CI-CD Refactor] Sample-outputs now Dict[str, NumpyDirectory] instead of NumpyDirectory

dbogunowicz:
Before:
we were assuming that the `ModelDirectory` contains a single `sample_output` folder.
Now:
`ModelDirectory` may contain several folders named `sample_output_{engine_name}`, which need to be handled by the API.

### DIFF
--- a/src/sparsezoo/v2/model_directory.py
+++ b/src/sparsezoo/v2/model_directory.py
@@ -552,7 +552,11 @@ class ModelDirectory(Directory):
         for directory in directories:
             engine_name = directory.name.split("_")[-1]
             if engine_name not in ENGINES:
-                raise ValueError("")
+                raise ValueError(
+                    f"The name of the 'sample_outputs' directory should "
+                    f"end with an engine name (one of the {ENGINES}). "
+                    f"However, the name is {directory.name}."
+                )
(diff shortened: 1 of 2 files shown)


src/sparsezoo/v2/model_directory.py
```diff
@@ -535,6 +544,22 @@ def _download(
...
             raise NotImplementedError()
 
+    @staticmethod
+    def _sample_outputs_list_to_dict(
+        directories: List[NumpyDirectory],
+    ) -> Dict[str, NumpyDirectory]:
+        engine_to_numpydir_map = {}
+        for directory in directories:
+            engine_name = directory.name.split("_")[-1]
+            if engine_name not in ENGINES:
```
bfineran, line 554: make sure that `torch`, `tensorflow_v1`, and `keras` are all included as valid engines

bfineran (commented):
looks good overall, need to make sure that original framework engines are supported (these are the source of truth for correctness)

src/sparsezoo/v2/model_directory.py
dbogunowicz: absolutely

PR #166, opened 2022-06-06
Feature/ci cd refactor/validations

dbogunowicz:


tests/sparsezoo/v2/test_model_directory.py
```diff
@@ -59,6 +60,7 @@
                 "sample_outputs",
                 "sample_labels",
```
dbogunowicz, line 61: Just making sure, the presence of `sample_labels` is also integration-dependent, or am I wrong here?

src/sparsezoo/v2/integration_validator.py
```diff
@@ -0,0 +1,220 @@
...
+                file_names = set(deployment_directory.get_file_names())
+                for optional_file in self.optional_training_files:
+                    file_names.discard(optional_file)
+                if expected_file_names != file_names:
+                    raise ValueError(
+                        f"Failed to find expected files "
+                        f"{expected_file_names.difference(file_names)} "
+                        f"in the `deployment` directory {deployment_directory.name}."
+                    )
+                return
```
KSGulin, line 151: This looks like a broken loop. Is this the intended behavior below? 
```suggestion
        for file in deployment_directory.files:
            if isinstance(file, Directory):
                raise ValueError(
                    "Found nested directories within `deployment` directory. "
                    "The directory may only contain files, not directories."
                )

          file_names = set(deployment_directory.get_file_names())
          for optional_file in self.optional_training_files:
              file_names.discard(optional_file)
          if expected_file_names != file_names:
              raise ValueError(
                  f"Failed to find expected files "
                  f"{expected_file_names.difference(file_names)} "
                  f"in the `deployment` directory {deployment_directory.name}."
                  )
          return
```

src/sparsezoo/v2/integration_validator.py
```diff
@@ -0,0 +1,220 @@
...
+                for optional_file in self.optional_training_files:
+                    file_names.discard(optional_file)
+
+                if expected_file_names != file_names:
+                    raise ValueError(
+                        f"Failed to find expected files "
+                        f"{expected_file_names.difference(file_names)} "
+                        f"in the `training` directory {training_directory.name}."
+                    )
+                return
```
KSGulin, line 183: See comment for `_validate_deployment_directory`

src/sparsezoo/v2/integration_validator.py
```diff
@@ -0,0 +1,220 @@
...
+        else:
+            # make sure that all default files in ModelDirectory are present
+            return [file for file in self.model_directory.files]
+
+    def _validate_deployment_directory(self, deployment_directory: File) -> None:
+        expected_file_names = self.integration_to_expected_training_files[
+            self.integration_name
+        ](deployment=True)
+        for file in deployment_directory.files:
+            if isinstance(file, Directory):
```
KSGulin, line 136: Should this check for list type as well?

src/sparsezoo/v2/integration_validator.py
```diff
@@ -0,0 +1,220 @@
...
+                        f"in the `deployment` directory {deployment_directory.name}."
+                    )
+                return
+
+    def _validate_training_directory(self, training_directory: File) -> None:
+        expected_file_names = self.integration_to_expected_training_files[
+            self.integration_name
+        ]()
+        for file in training_directory.files:
+            if isinstance(file, Directory):
```
KSGulin, line 158: See comment for` _validate_deployment_directory`

KSGulin (commented):
I like the move towards a self contained integration validator. Left a few comments.

src/sparsezoo/v2/integration_validator.py
```diff
@@ -0,0 +1,220 @@
...
+            )
+
+        for file in self.model_directory.files:
+            if isinstance(file, dict):
+                validations[file.__repr__()] = all(
+                    _file.validate() for _file in file.values()
+                )
+            elif isinstance(file, list):
+                validations[file.__repr__()] = all(_file.validate() for _file in file)
+            else:
```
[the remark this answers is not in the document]
dtransposed: Well, this would be ideally the case, but since `ModelDirectory` can only contain files which are either type: list, dict, File, or Directory, being super strict here is not necessary imo.

src/sparsezoo/v2/model_directory.py
```diff
@@ -230,55 +239,37 @@ def download(self, directory_path: str, override: bool = False) -> bool:
...
-                "If created using method `from_zoo_api`, "
-                "call `download()` method prior to `validate()`"
-            )
+        if validate_onnxruntime:
+            if not self.inference_runner.validate_with_onnx_runtime():
+                logging.warning(
+                    "Failed to validate the compatibility of "
+                    "`sample_inputs` files with the `model.onnx` model."
+                )
+                return False
```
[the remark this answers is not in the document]
dtransposed: Good point. This is actually an interesting engineering question. We might be raising warnings (that do not interrupt the code), or value errors (that do stop the script) or have a flag (e.g. `strict_mode = False`), that would allow to choose between both.
@bfineran do you have a opinion on this? Should failed validation just be flagged using logging warnings and the function should gracefully return `False`, or would we rather like to be strict and raise value errors anytime something goes south?

src/sparsezoo/v2/integration_validator.py
dbogunowicz: So the indentations are correct here, and so is the logic. However, it is not super easy to grasp what is happening at the first glance.

The point is to validate the training directory (the old framework folder). This directory can have two structures.

it contains checkpoint folders e.g checkpoint_01, checkpoint_02 etc. Inside the checkpoint folders, you would have training files e.g. tokenizer, model .pth, etc. (let's call it case 1)
it contains directly training files, there are no folders (case 2).
In case 2: we skip the if condition in L158 and jump directly to L172. We know that the first file in the training directory is a file, not a folder, so we know that this is case2. So we fetch the names of all the files in L173, compare them to the expected names and leave the loop. There is no need to loop through the other files, the first file already tells us everything we need.

In case1: we take the path starting at L157 and check every checkpoint folder, conveniently using recursion.

src/sparsezoo/v2/model_directory.py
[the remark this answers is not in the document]
dbogunowicz: Good point. This is actually an interesting engineering question. We might be raising warnings (that do not interrupt the code), or value errors (that do stop the script) or have a flag (e.g. strict_mode = False), that would allow to choose between both.
@bfineran do you have a opinion on this? Should failed validation just be flagged using logging warnings and the function should gracefully return False, or would we rather like to be strict and raise value errors anytime something goes south?

src/sparsezoo/v2/integration_validator.py
dbogunowicz: Answered in different thread, please resolve the conversation if it addresses the issue.

src/sparsezoo/v2/integration_validator.py
dbogunowicz: Answered in different thread, please resolve the conversation if it addresses the issue.

src/sparsezoo/v2/integration_validator.py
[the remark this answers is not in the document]
dbogunowicz: Well, this would be ideally the case, but since `ModelDirectory` can only contain files which are either type: list, dict, File, or Directory, being super strict here is not necessary imo.

src/sparsezoo/v2/integration_validator.py
dbogunowicz: The deployment directory is planned to be a non-nested directory, that only contains files of the type `File`. This if-statement is meant to catch the human error of e.g. pasting some folder into the `deployment` directory by mistake. This scenario is possible. However,  I do not really see how `file` in this context could have the type `List` or `Dict`.

src/sparsezoo/v2/integration_validator.py
dbogunowicz: "For case 1, is it expected that there may be both checkpoint folders and other files within the training directory?" -> nop, this should not happen.

Your proposal is good, I will merge it into my code, thanks.

src/sparsezoo/v2/integration_validator.py
```diff
@@ -0,0 +1,220 @@
...
+    :param model_directory: ModelDirectory class object to be
+        validated
+    """
+
+    def __init__(self, model_directory: ModelDirectory):
+
+        # mapping from the integration name to the helper function
+        # that outputs a set of expected `training` files
+        self.integration_to_expected_training_files = {
+            "nlp-question_answering": self._validate_nlp,
```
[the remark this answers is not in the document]
dbogunowicz: @bfineran not really sure what you mean by that. Currently, for every `ModelDirectory` we read out the integration data directly from the model card (domain + sub domain), and then we use the lookup table to relate the integration to the proper validation script.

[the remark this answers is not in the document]
dbogunowicz: Or would you rather have the user defining the validation files herself?

KSGulin (dismissed):
LGTM!

src/sparsezoo/v2/directory.py
```diff
@@ -27,6 +28,28 @@
 __all__ = ["Directory"]
 
 
+def _is_directory(file: File) -> bool:
```
bfineran, line 31: files don't need extensions, any reason not to use `os.path.isdir` instead?

src/sparsezoo/v2/directory.py
```diff
@@ -56,6 +81,22 @@ def __init__(
...
+    def from_file(cls, file: File) -> "Directory":
+        if not file.path:
+            raise ValueError(
+                f"Attempting to convert File class object {file.name} "
+                "to Directory class object, but the file.path is missing."
+            )
+        name = file.name
+        files = [
+            File(name=os.path.basename(path), path=path)
+            for path in glob.glob(os.path.join(file.path, "*"))
```
bfineran, line 94: any reason to not use `os.listdir` instead of this glob?

src/sparsezoo/v2/directory.py
```diff
@@ -157,6 +201,25 @@ def download(
...
+        :return: File if found, otherwise None
+        """
+        for file in self.files:
+            if file.name == file_name:
+                return file
+            if isinstance(file, Directory):
+                file = file.get_file(file_name=file_name)
+                if file:
+                    return file
+        logging.warning(f"File with name {file_name} not found!")
```
bfineran, line 220:
```suggestion
        logging.warning(f"File with name {file_name} not found")
```

src/sparsezoo/v2/integration_validation/integrations.py
```diff
@@ -0,0 +1,59 @@
...
+        "config.json",
+        "pytorch_model.bin",
+        "special_tokens_map.json",
+        "tokenizer.json",
+        "tokenizer_config.json",
+        "trainer_state.json",
+        "training_args.bin",
+        "vocab.txt",
+    }
+    optional_training_files = {"recipe.yaml"}
```
bfineran, line 32: `recipe.yaml` is mandatory for NLP. this is the most important file to validate for

src/sparsezoo/v2/integration_validation/integrations.py
```diff
@@ -0,0 +1,59 @@
...
+    training_files = {
+        "model.pth",
+    }
+    optional_training_files = {"recipe.yaml"}
+    additional_deployment_files = {"model.onnx"}
+    return training_files, optional_training_files, additional_deployment_files
+
+
+def validate_cv_detection():
+    training_files = {"model.pt", "model.ckpt.pt"}
```
bfineran, line 47: `model.ckpt.pt` is optional here

src/sparsezoo/v2/integration_validation/validator.py
```diff
@@ -0,0 +1,246 @@
...
+        successful
+    :param integration2data: mapping from the integration tag (string) from
+        ModelDirectory's model card to validation function
+    """
+
+    def __init__(
+        self,
+        model_directory: ModelDirectory,
+        essential_files: Set[str] = ESSENTIAL_FILES,
+        integration2data: Dict[str, Callable] = INTEGRATION_NAME_TO_DATA,
```
bfineran, line 64: nit: `integration_to_data` also consider different word than data if we're describing a function

src/sparsezoo/v2/integration_validation/validator.py
```diff
@@ -0,0 +1,246 @@
...
+        that need to be present, so that the validation is
+        successful
+    :param integration2data: mapping from the integration tag (string) from
+        ModelDirectory's model card to validation function
+    """
+
+    def __init__(
+        self,
+        model_directory: ModelDirectory,
+        essential_files: Set[str] = ESSENTIAL_FILES,
```
bfineran, line 63: nit: essiential -> `required`

src/sparsezoo/v2/integration_validation/validator.py
```diff
@@ -0,0 +1,246 @@
...
+            return all([file is not None for file in self.model_directory.files])
+
+    def _validate_deployment_directory(
+        self,
+        deployment_directory: File,
+        validation_files: Set,
+        optional_validation_files: Optional[Set] = set(),
+    ) -> None:
+        if any([isinstance(file, Directory) for file in deployment_directory.files]):
+            raise ValueError(
```
bfineran, line 174: is there any reason we don't allow for directories? None exist right now, but is there any reason we would want to block this in the future?

src/sparsezoo/v2/integration_validation/validator.py
```diff
@@ -0,0 +1,246 @@
...
+                    )
+                self._validate_training_directory(
+                    training_directory=file,
+                    validation_files=validation_files,
+                    optional_validation_files=optional_validation_files,
+                )
+        else:
+            # Training directory does not contain any subfolders,
+            # but the training files directly.
+            file_names = set(training_directory.get_file_names())
```
bfineran, line 215: looks like a good amount of this logic can be shared

src/sparsezoo/v2/model_directory.py
```diff
@@ -149,6 +157,9 @@ def __init__(
         )
 
         super().__init__(files=files, name=name, path=path, url=url)
+        from sparsezoo.v2.integration_validation.validator import IntegrationValidator
```
bfineran, line 160: why do we import here?, if necessary to do so at runtime, let's add a comment for why

bfineran (changes requested):

src/sparsezoo/v2/integration_validation/integrations.py
dbogunowicz: absolutely

src/sparsezoo/v2/integration_validation/integrations.py
dbogunowicz: absolutely

src/sparsezoo/v2/integration_validation/validator.py
dbogunowicz: this is the old logic I understood to be correct back then. will be replaced.

dbogunowicz:
@bfineran @KSGulin 
Addressed your comments. Highlights:
- short explanation why `IntegrationValidator` needs to be imported at runtime when constructing `ModelDirectory`
- simplified the logic for parsing `training` and `deployment` directories in `IntegrationValidator`
- found by accident a bug (nested directories were not parsed properly) and fixed it
- edited the validation data (this will probably still be evolving to some extent, but we can quickly accommodate for changes by simply changing the contents of the script, quite happy about it).

KSGulin (approved):

PR #171, opened 2022-06-14
[CI/CD Refactor] `setup_model_directory` helper function

dbogunowicz:
The goal of `setup_model_directory` is to consume the outputs from the integration pipeline and save them into one, neat directory to be further consumed to create the `ModelDirectory` class object.

### DIFF
--- a/src/sparsezoo/v2/helpers.py
+++ b/src/sparsezoo/v2/helpers.py
@@ -1,0 +1,76 @@
+"""
+Helper functions pertaining to the creation of ModelDirectory
+"""
+
+import copy
+from pathlib import Path
+from sparsezoo.v2 import File, Directory, NumpyDirectory, SampleOriginals
+from typing import Union, List
+import logging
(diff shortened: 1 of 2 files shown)


src/sparsezoo/v2/helpers.py
```diff
@@ -0,0 +1,120 @@
...
+    files_dict = copy.deepcopy(locals())
+    del files_dict["output_dir"]
+
+    for name, file in files_dict.items():
+        if file is None:
+            logging.debug(f"File {name} not provided, skipping...")
+        else:
+            if isinstance(file, str):
+                file = _create_file_from_path(file)
+            elif isinstance(file, list) and isinstance(file[0], str):
```
KSGulin, line 86: What is `isinstance(file[0], str)` checking for?

This feels like an instance where an else catch would be helpful

KSGulin (approved):

PR #176, opened 2022-06-29
[CI-CD Refactor] SparseZoo backend feature complete

dbogunowicz:
This PR is the final addition to the sparsezoo v2 codebase that adds a finishing touch to all the helper functions and objects revolving around `ModelDirectory`.

Contributions:
- an exhaustive set of unit tests
- hardened logic to handle `ModelDirectory` creation from `request.json` and local directory, as well as all class methods (special focus on the `download()` method).
- compatible with the latest agreements on the structure of `ModelDirectory`
- backward compatible with the directories that follow "v1 structure".
- removed the `SampleOriginals` class object, it became unnecessary over time and is replaced by `Directory`.


src/sparsezoo/v2/helpers.py
```diff
@@ -0,0 +1,200 @@
...
+        assert len(data) == 1
+        _, file_dict = data[0]
+        file_dict["file_type"] = "deployment"
+        request_json.append(file_dict)
+
+        # create files for onnx directory (use `model.onnx` to simulate them)
+        files_to_create = [
+            "model.11.onnx",
+            "model.14.onnx",
+        ]
```
bfineran, line 65: what is the goal here?

bfineran: why are we creating dummy files in `src` instead of `tests`?

src/sparsezoo/v2/inference/inference_runner.py
```diff
@@ -112,14 +113,12 @@ def validate_with_onnx_runtime(self) -> bool:
...
         for target_output, output in zip(sample_outputs, self._run_with_onnx_runtime()):
             target_output = list(target_output.values())
-            is_valid = [
-                numpy.allclose(o1, o2.flatten(), atol=1e-5)
-                for o1, o2 in zip(target_output, output)
-            ]
-            validation += is_valid
+            for o1, o2 in zip(target_output, output):
+                if o2.ndim != o1.ndim:
+                    o2 = o2.squeeze(0)
```
corey-nm, line 120: Is this to account for batch dimensions? Will there ever be an `o2` that has fewer than 1 dimension? ie. `o2.ndim > o1.ndim + 1` in which case the squeeze will still not make the ndims match

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -377,32 +395,56 @@ def _file_from_files(
...
-        if len(files_found) == 1:
+        elif len(files_found) == 1:
+            return files_found[0]
+
+        elif display_name == "model.onnx" and len(files_found) == 2:
+            # `model.onnx` file may be found twice:
+            #   - directly in the root directory
+            #   - inside `deployment` directory
+            # if this is the case, return the first file
+            # (the other one is a duplicate)
```
corey-nm, line 406: maybe add `assert files_found[0] == files_found[1]` to be sure?

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -377,32 +395,56 @@ def _file_from_files(
...
         # a Directory() object, if successful,
         # otherwise None.
-        directories_found = []
-        for file in files:
-            directory = self._get_directory(
-                file=file,
-                directory_class=directory_class,
-                display_name=display_name,
-                regex=regex,
+        if all([file_dict.get("file_type") for file_dict in files]):
```
corey-nm, line 423: is "check that all files have a file_type that is not None" the correct interpretation of this? guessing you can't do `"file_type" in file_dict` because you also want to make sure that `file_dict["file_type"] is not None`?

kylesayrs (approved):
Beautiful manipulation of paths. Ran tests on my computer

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -186,15 +176,19 @@ def from_directory(cls, directory_path: str) -> "ModelDirectory":
...
             raise ValueError(
                 "The directory path is empty. "
                 "Check whether the indicated directory exists."
             )
-        for path in paths:
-            display_name = os.path.basename(path)
-            files.append(file_dictionary(display_name=display_name, path=path))
+        for display_name in display_names:
+            files.append(
+                file_dictionary(
```
rahul-tuli, line 187: nit: This function is only used once, could we explicitly create the dictionary here?

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -238,9 +237,20 @@ def download(self, directory_path: str, override: bool = False) -> bool:
...
+                        raise ValueError(
+                            f"Attempted to download file {key}, " f"but it is `None`!"
+                        )
+
+                    else:
+                        logging.warning(
+                            f"Attempted to download file {key}, "
+                            f"but it is `None`. The file is being skipped..."
+                        )
         return all(downloads)
```
rahul-tuli, line 254: nit: again, can be ignored, but do we really need to create a `downloads` list in this case?

src/sparsezoo/v2/objects/model_directory.py
rahul-tuli: +1 for raising `ValueError` with check proposed by @corey-nm

src/sparsezoo/v2/objects/model_directory.py
rahul-tuli, line 423: Same old favorite nit to take advantage of `all`

```suggestion
        if all(file_dict.get("file_type") for file_dict in files):
```

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -422,25 +464,13 @@ def _download(
...
         if isinstance(file, File):
-            # TODO: Continuing with the hack
-            if file.name in SKIP_ATTRIBUTES:
-                return True
-
-            if file.url:
+            if file.url or (
+                file.url is None
+                and any(
+                    [isinstance(file, _class) for _class in [Directory, NumpyDirectory]]
```
rahul-tuli, line 471: Allow early exit for `any`
```suggestion
                    isinstance(file, _class) for _class in [Directory, NumpyDirectory]
```

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -454,7 +484,10 @@ def _download(
             return all(validations)
```
rahul-tuli, line 484: Naice

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -454,7 +484,10 @@ def _download(
...
 
         else:
-            raise NotImplementedError()
+            validations = (
+                self._download(_file, directory_path) for _file in file.values()
+            )
+            return all(validations)
```
rahul-tuli, line 490: Nice

src/sparsezoo/v2/utils/backwards_compatibility.py
```diff
@@ -0,0 +1,148 @@
...
+    training_file_names = [
+        x[1]["display_name"]
+        for x in fetch_from_request_json(request_json, "file_type", "training")
+    ]
+    nlp_folder = (
+        True
+        if (("config.json") in training_file_names)
+        and (("tokenizer.json") in training_file_names)
+        else False
+    )
```
rahul-tuli, line 79:
```suggestion
    nlp_folder = (
        "config.json" in training_file_names
        and "tokenizer.json" in training_file_names
    )
```

src/sparsezoo/v2/utils/backwards_compatibility.py
```diff
@@ -0,0 +1,148 @@
...
+    if nlp_folder:
+        for file_name in ["config.json", "tokenizer.json"]:
+            file_dict_training = fetch_from_request_json(
```
rahul-tuli, line 82: nit: Maybe rename `file_dict_training` to `training_file_dict`

src/sparsezoo/v2/utils/backwards_compatibility.py
```diff
@@ -0,0 +1,148 @@
...
+                request_json, "display_name", file_name
+            )
+            assert len(file_dict_training) == 1
+            file_dict_deployment = copy.copy(file_dict_training[0][1])
```
rahul-tuli, line 86: Same as above, nit: Maybe rename `file_dict_deployment` to `deployment_file_dict`

src/sparsezoo/v2/utils/backwards_compatibility.py
```diff
@@ -0,0 +1,148 @@
...
+            file_dict_deployment["file_type"] = "deployment"
+            request_json.append(file_dict_deployment)
+
+    # create recipes
+    recipe_dicts_list = fetch_from_request_json(request_json, "file_type", "recipe")
+    for (idx, file_dict) in recipe_dicts_list:
```
rahul-tuli, line 92: Does `recipe_dicts_list` always contain a tuple like `(index, file_dict)` or did we miss an `enumerate` here?

rahul-tuli (approved):

src/sparsezoo/v2/utils/backwards_compatibility.py
```diff
@@ -0,0 +1,148 @@
...
+    # searches through the `request_json` list to find a
+    # dictionary, that contains the requested
+    # key-value pair.
+    # return a list of tuples
+    # (every tuple is a file_dict, together
+    # with the respective list index)
+    return [
+        (idx, copy.copy(file_dict))
+        for (idx, file_dict) in enumerate(request_json)
+        if file_dict[key] == value
```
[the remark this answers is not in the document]
rahul-tuli: I have a suggestion, instead of using plain old tuples, maybe use `namedtuple` here, the idea being it works like a tuple, so your existing code will work + wherever we are using indexing we can replace that with named access

for example
```python
# using tuple
snorlax = ("Snorlax", 1_000_000)
weight = snorlax[1] # obscure index 1
```

with named tuple

```python
# with named tuple this becomes more readable

from collections import namedtuple
Something = namedtuple("Something", "name weight")

snorlax = Something(name='Snorlax', weight=1_000_000,)
weight = snorlax.weight # avoids obscure index
```

src/sparsezoo/v2/utils/backwards_compatibility.py
rahul-tuli: Also using named tuple as suggested here will avoid magic indexes

src/sparsezoo/v2/helpers.py
dbogunowicz: this has been resolved.

src/sparsezoo/v2/inference/inference_runner.py
dbogunowicz: Yes, this is to account for batch dimensions. Well, I am 99% sure that `o2.ndim` will always be `o1.ndim + 1` (this seems to be the case for all our integrations). Actually, I should probably remove the `if` statement here and apply the `squeeze` directly.

src/sparsezoo/v2/objects/directory.py
```diff
@@ -278,7 +280,11 @@ def _unpack(self):
         # To unpack the Directory the following criteria need to be fulfilled:
         # 1) The Directory needs to be a tar archive
         # 2) The Directory needs to have a `path` attribute.
-        return self.is_archive and self.path
+        return self.is_archive and self.path is not None
+
+    def __iter__(self) -> File:
```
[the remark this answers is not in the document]
dbogunowicz: image

src/sparsezoo/v2/inference/inference_runner.py
```diff
@@ -112,14 +113,12 @@ def validate_with_onnx_runtime(self) -> bool:
...
+                validation.append(numpy.allclose(o1, o2, atol=1e-5))
```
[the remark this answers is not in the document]
dbogunowicz: absolutely right. having the list was useful for debugging, but clearly too complicated.

src/sparsezoo/v2/objects/directory.py
```diff
@@ -287,11 +293,12 @@ def is_directory(file: File) -> bool:
...
-        raise ValueError(
-            "Cannot call the method. "
-            "The File/Directory class object's `path` "
-            "attribute is None."
-        )
+        from pathlib import Path
+
+        # we are processing a downladable file
+        file_name_without_extension = Path(file.name).stem
+        return file_name_without_extension == file.name
```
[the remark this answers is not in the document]
dbogunowicz: @rahul-tuli precisely. 

```
 if file.path is None: # file.path is None --> file is does not have a local path (only url to be downloaded)
        # we are processing a downloadable file
        file_name_without_extension = pathlib.Path(file.name).stem 
        return file_name_without_extension == file.name
```

src/sparsezoo/v2/objects/model_directory.py
dbogunowicz: Well, `ValueError` here is actually pretty useless. `files_found[0] == files_found[1]` should be just the condition of the `if` statement.

dbogunowicz: After taking a second look, I do not think that this assertion is really needed.

src/sparsezoo/v2/objects/directory.py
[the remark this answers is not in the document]
dbogunowicz: Lemme resolve the conflict: https://docs.python.org/3/library/typing.html

image

src/sparsezoo/v2/objects/file.py
```diff
@@ -120,26 +120,26 @@ def download(
...
-                    self.path = new_file_path
-                    return
-
-                except Exception as err:
-                    logging.error(err)
-                    logging.error(traceback.format_exc())
-                    time.sleep(retry_sleep_sec)
-                logging.error(
-                    f"Trying attempt {attempt + 1} of {retries}.", attempt + 1, retries
+        for attempt in range(retries):
```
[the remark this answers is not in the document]
dbogunowicz: Why? Looks good to me.

src/sparsezoo/v2/objects/model_directory.py
dbogunowicz: No no, all I want is to make sure that `file_dict` has a key `file_type`. So thanks, your proposal is actually much better (less ambiguous) than what I have implemented.

src/sparsezoo/v2/objects/model_directory.py
dbogunowicz: good idea!

dbogunowicz: Actually, this factory method will be deprecated in the next PR, so let's drop this for now. But I will take note to use your suggestion in the other function which will use the same logic ok?

src/sparsezoo/v2/objects/model_directory.py
dbogunowicz: I still like it being there for the debug properties, to be honest.

src/sparsezoo/v2/utils/backwards_compatibility.py
dbogunowicz: Good one!

dbogunowicz: I like that even more: `{"config.json", "tokenizer.json"}.issubset(set(training_file_names))`.
I've become  a set junkie lately.

src/sparsezoo/v2/utils/backwards_compatibility.py
dbogunowicz: Yes, it will. But I will use your `namedTuple` suggestion I think.

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -87,80 +83,76 @@ def __init__(
...
-            self.logs,
-            self.onnx_folder,
-            self.onnx_model,
-            self.analysis,
-            self.benchmarks,
-            self.eval_results,
-            self.model_card,
-            self.recipes,
-        ]
+        self._files_dictionary = {
```
corey-nm, line 126: What's the reasoning behind the change from list to dictionary? Is the `key` here better than using  file/directory `.name`?

src/sparsezoo/v2/objects/model_directory.py
```diff
@@ -60,13 +60,11 @@ def __init__(
     ):
 
         self.training: Directory = self._directory_from_files(
```
corey-nm, line 62: How would you feel about something like this instead of manually calling `*_from_files()` for each of the expected files/directories? I think this could 1. make it a bit easier to quickly parse exactly what is expected to be in the directory, and 2. reduce chance for copy/paste error

```python
structure = {
    "sample_originals": Directory,
    "training": Directory,
    "sample_inputs": NumpyDirectory,
    "model_card": (File, "model.md"),
    ...
}

files = {}
for key, value in structure.items():
    if isinstance(value, tuple):
        ty, display_name = value
    else:
        ty, display_name = value, key
    
    if issubclass(ty, Directory):
        files[key] = self._directory_from_files(files, directory_class=ty, display_name=display_name)
    elif issubclass(ty, File):
        files[key] = self._file_from_files(files, display_name=display_name)
```

And then if files was just a list (or the dictionary key was exactly the display name), you could get rid of the tuple value and just go `{<display_name>: Type}`

corey-nm: (And it makes the ones that don't fit into that structure a bit more obvious - I think just `sample_outputs`?)

src/sparsezoo/v2/objects/directory.py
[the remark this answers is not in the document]
kylesayrs: Hm technically this `is_directory` function won't work if a file without an extension is added such as Makefile, LICENSE, ect. Perhaps there should be a warning about this failure in the contract.

src/sparsezoo/v2/objects/model_directory.py
dbogunowicz: Yeah, this is to easy iterate over files for download in `L242` (have name for messages/errors together with the file).

src/sparsezoo/v2/objects/file.py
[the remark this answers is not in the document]
bfineran: not sure why github isn't syntax highlighting this file...

bfineran (approved):

PR #177, opened 2022-07-19
Updated supported python versions in setup.py

rahul-tuli:
Update supported Python versions in `setup.py` for better error message

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,7 +123,7 @@ setup(
     install_requires=_setup_install_requires(),
     extras_require=_setup_extras(),
     entry_points=_setup_entry_points(),
-    python_requires=">=3.6.0",
+    python_requires=">=3.6.0,<3.10",
     classifiers=[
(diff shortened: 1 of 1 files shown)


corey-nm:
Wait does sparsezoo actually have limits on python version? If not we should close this one

dbarbuzzi:
We currently [support Python 3.10 in deepsparse](https://github.com/neuralmagic/deepsparse/blob/main/setup.py#L275) (which depends on sparsezoo) – wouldn’t this break that support?

corey-nm:
> We currently [support Python 3.10 in deepsparse](https://github.com/neuralmagic/deepsparse/blob/main/setup.py#L275) (which depends on sparsezoo) – wouldn’t this break that support?

Yeah good point. I'm thinking we can just close this PR since I don't think sparsezoo has any limits on python version. Sound reasonable?

PR #179, opened 2022-07-20
[CI-CD Refactor] Sparsezoo User API

dbogunowicz:
Adds the pythonic user API for the SparseZoo.
What is still missing are small features, that I plan to implement along the way:
- [CI-CD Refactor] Add custom __str__() method to .md files (recipes, model cards) #182
- [CI-CD Refactor] Add proper hashing function to create model name in ~/.cache/sparsezoo #181

# Testing with no string arguments

```python
## Directory from API
model = Model(path=stub)
deployments = model.deployment.available
default_deployment = model.recipes.default
default_deployment = model.deployment["default"]
default_deployment_path = default_deployment.get_path()
print(f"Available deployments: {deployments}. Default deployment: '{default_deployment}', with path: {default_deployment_path}")
```
```bash
Available deployments: ['default']. Default deployment: '<sparsezoo.v2.objects.model_objects.SelectDirectory object at 0x7fe7c718fb20>', with path: /home/damian/.cache/sparsezoo/T8JLQ9/deployment
```

```python
## Checkpoint from API
model = Model(stub)
checkpoints = model.training.available
default_checkpoint = model.training.default
default_checkpoint = model.training["preqat"]
default_checkpoint_path = default_checkpoint.get_path()
print(f"Available checkpoints: {checkpoints}. Default checkpoint: '{default_checkpoint}', with path: {default_checkpoint_path}")
```
```bash
Available checkpoints: ['preqat']. Default checkpoint: '<sparsezoo.v2.objects.model_objects.SelectDirectory object at 0x7f69df0018e0>', with path: /home/damian/.cache/sparsezoo/I23IGU/training
```

```python
## Recipes from API
model = Model(stub)
recipes = model.recipes.available
chosen_recipe = model.recipes["transfer_learn"]
default_recipe = model.recipes.default
print(f"Available checkpoints: {recipes}.")
print(f"Default recipe name is : '{default_recipe.name}', with path: {default_recipe.get_path()}")
print(f"Chosen recipe name is : '{chosen_recipe.name}', with path: {chosen_recipe.get_path()}")
```
```bash
Available recipes: ['original', 'transfer_learn'].
Default recipe name is : 'recipe_original.md', with path: /home/damian/.cache/sparsezoo/6VZWVB/recipe/recipe_original.md
Chosen recipe name is : 'recipe_transfer_learn.md', with path: /home/damian/.cache/sparsezoo/6VZWVB/recipe/recipe_transfer_learn.md
```

# Testing more complex functionalities

```python
stub = stub + "?" + "recipe=transfer_learn" + "&" + "checkpoint=preqat"
model = Model(path=stub)
recipes_path = model.recipes.get_path()
print(f"Recipes folder is located at {recipes_path}.")
# accessing wrong recipe
recipes_available = model.recipes.available
print(f"Available recipes: {recipes_available}")
recipe_transfer_learn = model.recipes["transfer_learn"]
print(f"Transfer learn recipe has path: {recipe_transfer_learn.get_path()}")
# >> model.recipes["original"] will give you key error
recipe = model.recipes["transfer_learn"]

## Let's also show that we can download all the files

# there is already `recipe` folder in the model directory, so we need to
# set override to True to download all model files.
model.download(override=True)
model_path = model.get_path()
print(f'Model is located at {model_path}')
print(f'Inside we find files {os.listdir(model_path)}')
```
```bash
Recipes folder is located at /home/damian/.cache/sparsezoo/1BDKHB/recipe.
Available recipes: ['transfer_learn']
Transfer learn recipe has path: /home/damian/.cache/sparsezoo/1BDKHB/recipe/recipe_transfer_learn.md

downloading...: 100%|█████████| 17038995/17038995 [00:01<00:00, 12583419.35it/s]
downloading...: 100%|█████████| 17033252/17033252 [00:01<00:00, 12468182.70it/s]
downloading...: 100%|███████████| 4263560/4263560 [00:00<00:00, 12081952.31it/s]
downloading...: 100%|███████████| 3474675/3474675 [00:00<00:00, 11039627.84it/s]
downloading...: 100%|██████████████| 154707/154707 [00:00<00:00, 2318716.55it/s]
downloading...: 100%|██████████████| 174907/174907 [00:00<00:00, 2970562.68it/s]
downloading...: 100%|██████████████████| 2110/2110 [00:00<00:00, 2507076.90it/s]
WARNING:root:Overwriting the current location of the File: /home/damian/.cache/sparsezoo/1BDKHB/recipe/recipe_transfer_learn.md with the new location: /home/damian/.cache/sparsezoo/1BDKHB/recipe/recipe_transfer_learn.md.
downloading...: 100%|██████████████████| 2538/2538 [00:00<00:00, 1560871.49it/s]
downloading...: 100%|█████████| 17033252/17033252 [00:01<00:00, 12411867.87it/s]

Model is located at /home/damian/.cache/sparsezoo/L271EB
Inside we find files ['sample_labels.tar.gz', 'training', 'model.onnx', 'deployment', 'sample_originals.tar.gz', 'model.md', 'sample_outputs.tar.gz', 'recipe', 'sample_inputs.tar.gz']
```


src/sparsezoo/v2/utils/authentication.py
```diff
@@ -0,0 +1,183 @@
...
+    path = clean_path(path)
+
+    try:
+        os.makedirs(path)
+    except OSError as e:
+        if e.errno == errno.EEXIST:
+            pass
+        else:
+            # Unexpected OSError, re-raise.
+            raise
```
corey-nm, line 54:
```suggestion
    os.makedirs(path, exist_ok=True)
```

src/sparsezoo/v2/objects/model_objects.py
```diff
@@ -114,3 +119,73 @@ def _validate_model(self, model: onnx.ModelProto) -> bool:
...
+        )
+
+        self.files_dict = self.files_to_dictionary()
+
+    def __getitem__(self, key):
+        file = self.files_dict[key]
+        file.get_path()
+        return file
+
+    def files_to_dictionary(self):
```
corey-nm, line 158: Recommend just putting this code directly in `__init__` if its not called anywhere else (doesn't seem like it is?)

src/sparsezoo/v2/objects/model_objects.py
```diff
@@ -114,3 +119,73 @@ def _validate_model(self, model: onnx.ModelProto) -> bool:
...
+        super().__init__(
+            files=files, name=name, path=path, url=url, owner_path=owner_path
+        )
+
+        self.files_dict = self.files_to_dictionary()
+
+    def __getitem__(self, key):
+        file = self.files_dict[key]
+        file.get_path()
+        return file
```
corey-nm, line 156: So the only thing this SelectDirectory has over Directory is this `__getitem__`? Could this be added to Directory instead of adding a new class?

src/sparsezoo/v2/utils/authentication.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+CREDENTIALS_YAML = os.path.abspath(
+    os.getenv("SPARSEZOO_CREDENTIALS")
+    if os.getenv("SPARSEZOO_CREDENTIALS")
+    else clean_path(os.path.join("~", ".cache", "sparsezoo", "credentials.yaml"))
+)
+
+
+class SparseZooCredentials:
```
corey-nm, line 80: I think this could be reworked into two private_functions:
1. _save_token (unchanged from the method, but just a standalone function)
2. _get_token

The `_get_token` function could combine the `__init__` function and the `token` property so all the invariants are in one place.

Something like:

```python
def _get_token():
    if not os.path.exists(CREDENTIALS_YAML):
        return None
    with open(...) as fp:
        creds = yaml.safe_load(...)
    if creds is None or TOKEN_KEY not in creds:
       return None
    token = creds[TOKEN_KEY]["token"]
    created = creds[TOKEN_KEY]["created"]

    creation_difference = ...
    if creation_difference > 30:
        return None

    return token
```

then in get_auth_header you could just call `_get_token` in place of instantiating this class and access the token property

src/sparsezoo/v2/utils/model_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+__all__ = ["load_files_from_stub", "load_files_from_directory", "generate_model_name"]
+
+_LOGGER = logging.getLogger(__name__)
+
+ZOO_STUB_PREFIX = "zoo:"
+
+
+def file_dictionary(**kwargs):
+    return kwargs
```
corey-nm, line 34: Maybe delete this function and use `dict()` instead? see comment below

src/sparsezoo/v2/utils/model_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    display_names = os.listdir(directory_path)
+    if not display_names:
+        raise ValueError(
+            "The directory path is empty. "
+            "Check whether the indicated directory exists."
+        )
+    files = [
+        file_dictionary(
+            display_name=display_name, path=os.path.join(directory_path, display_name)
+        )
```
corey-nm, line 52:
```suggestion
        dict(
            display_name=display_name, path=os.path.join(directory_path, display_name)
        )
```

src/sparsezoo/v2/objects/model.py
```diff
@@ -215,23 +207,36 @@ def generate_outputs(
             yield output
 
     def download(
```
corey-nm, line 209: Thoughts on making this the main function for getting the model path in place of `get_path`? This could just return the model path instead of a bool:

```diff
-model.download(override=True)
-model_path = model.get_path()
+model_path = model.download(override=True)
```

corey-nm:
Question: are `model.deployment` and `model.training` both directories that contain checkpoints? If so what are thoughts on `model.checkpoints.deployment` and `model.checkpoints.training`?

src/sparsezoo/v2/objects/model.py
corey-nm: Actually following up with this - could users just do `Model("zoo:...", download=True)`? so its just one line for them?

src/sparsezoo/v2/utils/model_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    return stub, params
+
+
+# TODO: Remove that function later
+def generate_model_name(size=6, chars=string.ascii_uppercase + string.digits):
+    """
+    Create simple randomized string that can temporarily serve as a hash name
+    for the model
+    """
+    return "".join(random.choice(chars) for _ in range(size))
```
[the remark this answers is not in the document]
dbogunowicz: Good one.

src/sparsezoo/v2/objects/file.py
```diff
@@ -71,29 +78,43 @@ def __init__(
...
         url = file.get("url")
 
-        return cls(
-            name=name,
-            path=path,
-            url=url,
-        )
+        return cls(name=name, path=path, url=url, owner_path=owner_path)
+
+    def get_path(self, download_directory: Optional[str] = None):
```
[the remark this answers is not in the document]
dbogunowicz: This would be a bit ugly in terms of user experience. The objective of `get_path()` is to be seamlessly evoked directly as a method of a class object. Your suggestion implies additionally exporting the `get_path` function, which is too much of a hassle IMO.

src/sparsezoo/v2/objects/model_objects.py
dbogunowicz: Well, not only. `SelectDirectory` basically extends the `Directory` with the possibility to 
- select default subdirectories through the `default` property or "dict-like" key selection
- looking up available subdirectories through `available` property.

I feel that those two classes are distinct enough so that their separate existence is justified. `SelectDirectory` basically is a special, advanced case of `Directory`. I'm confident that not all directories should be `SelectDirectories`.

src/sparsezoo/v2/objects/model_objects.py
dbogunowicz: Very good idea.

src/sparsezoo/v2/utils/authentication.py
dbogunowicz: Good catch. Just for context, there is a bunch of functions that I am directly copying-pasting from sparsezoo v1; this one included. Once again, I propose to leave those as they are for now, and do a proper boy-scout refactoring before they land in main. Open for discussion.

src/sparsezoo/v2/utils/model_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    for file_dict in files:
+        if "recipe" in available_params and file_dict["file_type"] == "recipe":
+            value = params["recipe"]
+            if file_dict["display_name"] != "recipe_" + value + ".md":
+                continue
+        if "checkpoint" in available_params and file_dict["file_type"] == "training":
+            pass
+
+        if "deployment" in available_params and file_dict["file_type"] == "deployment":
+            pass
```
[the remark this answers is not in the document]
dbogunowicz: Yup, those are the placeholders for the future. Once the server migration is ready, we will be receiving API responses in a new format, that will be handled by those statements.

src/sparsezoo/v2/objects/model_objects.py
dbogunowicz: On the other hand, I like the aesthetics of not overloading the constructor with lines of code. Keeping it as a method, even though it's only called once, is just more readable I think.

src/sparsezoo/v2/objects/model.py
dbogunowicz: The first suggestion is very interesting, I like it. What do you think @bfineran ?
The second suggestion is probably not what I'd like to achieve. We are pretty convinced that `Model`'s `__init__` should be a "monadic" method. The user should just enter the `path` to the model and not care about any additional flags.

dbogunowicz:
@corey-nm 
`Deployment` and `training` are very distinct beasts. They would contain a slightly different set of files and be consumed by different consumers.
To explain it quickly:
- `training` is what SparseML will be using. For E.g. you want to continue training from a checkpoint. The `Training` folder will contain different checkpoints. You can grab one of the checkpoints - it contains all the files needed to resume training from this specific state - and continue training. You can expect `.pth` files to be located in checkpoints.
- `deployment` is what DeepSparse will be using. This is a "mature" version of the model, prepared and optimized for inference. You will never use the contents of `deployment` to e.g. continue training. Only to run inference. Therefore you can expect e.g. `.onnx` files to reside there.

corey-nm:
@dbogunowicz great descriptions thanks!! Are the names of those flexible? Something like `model.checkpoints` and `model.onnx` seem to better reflect the contents based on your descriptions

src/sparsezoo/v2/objects/model_objects.py
corey-nm: Ahh yeah, esp the default property since that specializes the directory structure to contain very specific things.

Thanks for the clarifications!

src/sparsezoo/search.py
```diff
@@ -0,0 +1,130 @@
...
+        sub_architecture = '-' + sub_architecture
+
+    stub = f"zoo:{domain if domain is not None else ''}/" \
+           f"{sub_domain if sub_domain is not None else ''}/" \
+           f"{architecture if architecture is not None else ''}" \
+           f"{sub_architecture if sub_architecture is not None else ''}/" \
+           f"{framework if framework is not None else ''}/" \
+           f"{repo if repo is not None else ''}/" \
+           f"{dataset if dataset is not None else ''}/" \
+           f"{sparse_tag if sparse_tag is not None else ''}"
```
corey-nm, line 128: So if a bunch of these are none would it just have a lot of slashes next to each other?

`zoo://////`?

src/sparsezoo/search.py
```diff
@@ -0,0 +1,130 @@
...
+    dataset = model_dict.get("dataset")
+    sparse_tag = model_dict.get("sparse_tag")
+
+    if sub_architecture is not None:
+        sub_architecture = '-' + sub_architecture
+
+    stub = f"zoo:{domain if domain is not None else ''}/" \
+           f"{sub_domain if sub_domain is not None else ''}/" \
+           f"{architecture if architecture is not None else ''}" \
+           f"{sub_architecture if sub_architecture is not None else ''}/" \
```
corey-nm, line 124: is it possible to have architecture None and sub_architecture not None? `"/-<sub arc>/"`

kylesayrs (dismissed):
I ran into an issue when trying to install with dev dependencies

`python3 -m pip install -e .[dev]`
`>>>FileNotFoundError: [Errno 2] No such file or directory: 'src/sparsezoo/version.py'`

dbogunowicz:
@kylesayrs could you try again in a fresh environment? I am reproducing it, but not getting the error.

src/sparsezoo/search.py
dbogunowicz: Not sure, probably not.

src/sparsezoo/search.py
dbogunowicz: Technically yes.

kylesayrs:
@dbogunowicz Very weird that you wouldn't be getting the error, maybe I'm missing something. I've tried with fresh venvs across two machines and it's the same missing file error. It's caused by this line in `setup.py`: `exec(open(os.path.join("src", "sparsezoo", "version.py")).read())`

src/sparsezoo/inference/inference_runner.py
```diff
@@ -97,9 +100,6 @@ def generate_outputs(
                 f"Supported engines: {self.engine_type_to_iterator.keys()}."
             )
 
-        if save_to_tar:
-            self._save_outputs_to_tar(iterator, engine_type)
-
```
corey-nm, line 102: Does this functionality exist somewhere else now?

src/sparsezoo/objects/utils.py
```diff
@@ -11,25 +11,198 @@
...
+    training_file_names = [
+        x[1]["display_name"]
+        for x in fetch_from_request_json(request_json, "file_type", "training")
+    ]
+    nlp_folder = (
+        True
+        if (("config.json") in training_file_names)
+        and (("tokenizer.json") in training_file_names)
+        else False
+    )
```
corey-nm, line 133:
```suggestion
    nlp_folder = ("config.json" in training_file_names) and ("tokenizer.json" in training_file_names)
```

src/sparsezoo/search.py
corey-nm: Ahh I see these values are from the backend, so we expect them to be well formed?

src/sparsezoo/utils/numpy.py
```diff
@@ -104,13 +63,35 @@ def save_numpy(
     return export_path
 
 
+def _fix_loaded_numpy(array) -> Union[numpy.ndarray, Dict[str, numpy.ndarray]]:
+    if not isinstance(array, numpy.ndarray):
```
corey-nm, line 67: Is this block assuming array is a `Dict`? should it checkout for isinstance(Dict)?

bfineran (changes requested):
the hard work shows on the updated python API @dbogunowicz. biggest thing to resolve is restoring the functionality/files that were mistakenly deleted in the refactor

src/sparsezoo/utils/numpy.py
```diff
@@ -26,45 +21,10 @@
...
-    Load a numpy file into either an ndarray or an OrderedDict representing what
-    was in the npz file
-    :param file_path: the file_path to load
-    :return: the loaded values from the file
-    """
-    file_path = clean_path(file_path)
-    array = numpy.load(file_path)
-
-    return _fix_loaded_numpy(array)
+__all__ = ["save_numpy", "load_numpy_list"]
```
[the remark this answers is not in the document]
dbogunowicz: Yes, I have salvaged the whole `numpy.py` script, I am a huge fan of how useful this set of helper functions is.

src/sparsezoo/objects/model_objects.py
```diff
@@ -23,13 +23,15 @@
 
 import onnx
```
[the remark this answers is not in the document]
dbogunowicz: I chose `directories` for lack of a better idea. If you have any proposal, feel free to drop a comment below.

src/sparsezoo/utils/numpy.py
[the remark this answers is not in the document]
dbogunowicz: Is this what you are referring to?

src/sparsezoo/search.py
dbogunowicz: Yes! :)

src/sparsezoo/objects/model_helpers.py
```diff
@@ -0,0 +1,159 @@
...
+    for file_dict in files:
+        if "recipe" in available_params and file_dict["file_type"] == "recipe":
+            value = params["recipe"]
+            if file_dict["display_name"] != "recipe_" + value + ".md":
+                continue
+        if "checkpoint" in available_params and file_dict["file_type"] == "training":
+            pass
+
+        if "deployment" in available_params and file_dict["file_type"] == "deployment":
+            pass
```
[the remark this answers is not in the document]
dbogunowicz: In theory, we refrain from adding todos. There will be an explicit need to revisit those functionalities once the server migration is complete.

src/sparsezoo/inference/inference_runner.py
dbogunowicz: Yes. It lives now in `sparsezoo/src/sparsezoo/objects/utils.py :: save_outputs_to_tar`

src/sparsezoo/objects/file.py
```diff
@@ -12,367 +12,304 @@
...
+            result = regex.search(yaml_str)
+            yaml_str = result.group(1)
+            yaml_dict = yaml.safe_load(yaml_str)
+            # returns a string "{domain}-{sub_domain}, if valid
+            # this makes the method reusable to fetch the integration
+            # name for the integration validation
+            return yaml_dict
+
+        except Exception as error:  # noqa: F841
+            logging.error(error)
```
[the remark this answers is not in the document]
dbogunowicz: Absolutely! Thanks for catching this.

src/sparsezoo/objects/file.py
```diff
@@ -12,367 +12,304 @@
...
+        # if not, attempt to check if it is a recipe file
+        if not is_valid_model_card:
+            try:
+                from sparseml.pytorch.optim import (  # noqa  F401
+                    ScheduledModifierManager,
+                )
+            except Exception as error:  # noqa  F841
+                # if not model card and unable to check if recipe,
+                # optimistically assume the .md file is valid.
+                return
```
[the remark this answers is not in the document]
dbogunowicz: In this scenario no. So what happens:

if the `.md` file is not a model card, we may attempt to validate the recipe file (also `.md` file):
- we do not expect `sparseml` to be installed by default. A developer can do this to be able to run `self._validate_recipe()`, but that's optional.
- if `sparseml` not installed (default scenario), simply ignore the `self._validate_recipe()` and gracefully continue with the logic of the code.

dbogunowicz:
@kylesayrs I need to admit that I was wrong. You were absolutely right, the `version.py` is missing and that is a huge no-go. I will shortly post another PR that patches this issue. Please ignore it in the context of this PR.

src/sparsezoo/objects/file.py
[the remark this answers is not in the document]
dbogunowicz: It should not, otherwise we would never have the opportunity to check whether the .md file is a recipe (when we fail to confirm that it is a model card)

src/sparsezoo/objects/model_helpers.py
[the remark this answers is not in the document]
corey-nm: Huh why refrain from adding todos? Is it captured somewhere else? Otherwise we have to keep this info in our heads and people who don't know will be confused when they come across these blocks.

src/sparsezoo/inference/inference_runner.py
corey-nm: Does this method still need to call that function though? Is `save_to_tar` flag not supported anymore?

src/sparsezoo/objects/file.py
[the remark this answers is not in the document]
corey-nm: Could you change this function to just be:
```python 
if not self._validate_model_card():
    self._validate_recipe(...)
```

And then do the import testing in _validate_recipe() so thats all in one place?

src/sparsezoo/objects/model_helpers.py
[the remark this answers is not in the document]
dbogunowicz: Hmm. Since I we are merging this PR into feature branch and not main, maybe adding a todo here to have a "mental marker" is actually reasonable. @bfineran ?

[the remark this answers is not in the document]
dbogunowicz: @corey-nm the general rule is no todos in the production code.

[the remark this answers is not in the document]
corey-nm: How about `NOTE: these will be filled out in future releases and are left here as a reminder`? 😁

src/sparsezoo/objects/file.py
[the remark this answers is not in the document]
dbogunowicz: yeah :D

src/sparsezoo/model/model.py
```diff
@@ -281,19 +282,50 @@ def validate(
...
+                "Model card missing! Before running `validate()` "
                 "method, the respective files must be present locally. "
-                "The solution is to call `download()` first."
+                "The solution may be to call `download()` first."
             )
 
         return self.integration_validator.validate(minimal_validation)
 
     def analyze(self):
         raise NotImplementedError()
```
corey-nm, line 295: This may be outside scope of this PR, but what's this function for? Could it just be deleted?

src/sparsezoo/model/model.py
```diff
@@ -15,52 +15,72 @@
...
+        a) a SparseZoo model stub:
+            i) without additional string arguments
+                e.g. 'zoo:model/stub/path'
+            ii) with additional string arguments (that will restrict
+                the set of file/directory files handled by this
+                `Model` object)
+                e.g. 'zoo:model/stub/path?param1=value1&param2=value2'
+        b) a local directory path
+            e.g. `/home/user/model_path`
+    """
```
corey-nm, line 67: This class is super easy to use from python interpreter. Maybe just add something to docstring about which methods/propeties should be used? Seems like generate_outputs() and maybe path are main ones?

bfineran (approved):

PR #180, opened 2022-07-20
Feature/ci cd refactor/user journey

dbogunowicz:


src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
+# API User pathway
+
+## Download one file
+
+### Through the string arguments
```
bfineran, line 5: IMO the string query param arguments shouldn't restrict `model.download()` functionality to only grab those files. instead, should just filter the relevant ones. (ie if recipe=transfer_learn, then `model.download` downloads everything but for checkpoints and recipes only includes the transfer learn ones)

for downloading specific directories/files - we can filter via the query params that were passed in

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
...
+...
+```
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+stub = stub + "?" + "deployment" # if we want the whole directory, the parameter does not have to have value
+model = Model(path=stub)
+model.download()
+deployment: str = model.recipes.downloaded_path()
```
bfineran, line 23: similarly to the v1 api let's make `downloaded_path` call `download()` if the file is not downloaded yet, that way we can save the extra call to `downlaod()` on line 22

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
...
+deployment: str = model.recipes.downloaded_path() 
+...
+```
+### Through the API purely
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+model = Model(path=stub)
+recipe: str = model.recipes.download(recipe="transfer_learn")
```
bfineran, line 32: how will a user know which recipes are available?

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
...
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+model = Model(path=stub)
+recipe: str = model.deployment.download()
+```
+
+## Download multiple chosen files (also in parallel)
+
+### Through the string arguments
```
bfineran, line 45: similar to above, the request for which files to be downloaded should be done through the python api, the query args should just be for filtering out specific kinds of files

ie

```
stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
# choose the files through the string arguments
stub = stub + "?" + "recipe=transfer_learn" + "&" + "checkpoint=preqat" + "&" + "recipe=original"

model = Model(stub)
model.download()
```
would download specified files + the sample inputs/outputs etc
and 
```python
model.training.downloaded_path()
```
for the same model would just download the preqat training checkpoint

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
...
+model = Model(path=stub)
+model.download()
+training: str = model.training.downloaded_path() 
+deployment: str = model.deployment.downloaded_path()
+model_card: str = model.model_card.downloaded_path()
+...
+```
+## QnA
+Q: What happens if user requests a file/directory by the invalid string param 
+A: Error on Model class construction, that says that the param is invalid and proposes the valid params in the error message
```
bfineran, line 86: 👍

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
...
+
+Q: What happens if the user calls `.download()` on the file that fails to get downloaded
+A: Verbose error, that either explains why 
+- a file cannot be downloaded (e.g. if the user calls `.download()` on the file that we do not have data on from the API)
+- a file that should be downloaded does not get downloaded (e.g. if timeout on waiting from the response from the server)
+
+Q: What happens if the user calls `.downloaded_path()` on a file that does not have one
+A: Verbose error, that explains why one cannot call `.downloaded_path()` and shortly proposes a solution (if exists)
```
bfineran, line 94: see comment about downloaded_path above

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Questions from my side:
+1. Should the user be able to download and access any file? I am talking about "private" ones like `benchmark.yaml`, `eval.yaml`,
```
bfineran, line 97: I don't think there are any "private files" `model.download` should download all for the most part (need to think about if we want to include all logs and checkpoints as that can get heavy)

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,100 @@
...
+`sample_outputs` or  `logs`.
+2. Should the user be able to download each and any files using string parameters? Or just `recipe`, `checkpoint` and `deployment`.
```
bfineran, line 99: let's check with @markurtz on the state of where we want query params, this needs to line up with the backend API so we should design carefully here and not assume too much

bfineran (changes requested):

src/sparsezoo/v2/test.md
dbogunowicz: ok, sounds good.

src/sparsezoo/v2/test.md
dbogunowicz: good idea

src/sparsezoo/v2/test.md
dbogunowicz: got it

src/sparsezoo/v2/test.md
dbogunowicz: Precisely my point. The long tail of users will not be needing them. Especially the `onnx` file may be the biggest burden on the download time (several onnx files, perhaps used by a negligibly small fraction of users).

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,102 @@
...
+model.recipes.download(recipe="transfer_learn")
+recipe: str = model.recipes.downloaded_path()
+```
+
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+model = Model(path=stub)
+recipe: str = model.deployment.download()
```
dbogunowicz, line 41:
```suggestion
deployment: str = model.deployment.download()
```

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,111 @@
...
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+stub = stub + "?" + "recipe=transfer_learn"
+# will read all the files from the server, except all the recipe files. The only recipe 
+# file that will be "seen" by the `model` is "transfer learn"
+model = Model(path=stub)
+model.download() # optional: may be called to download all the files, but is not required (see below). Will be omitted in further examples
+recipes: str = model.recipes.downloaded_path() # will automatically download the file (if not already downloaded) and return the path
```
markurtz, line 15: rewrite downloaded_path to path so it is more generic

additionally, recipes should be a parent folder that has a __dict__ implementation for individual recipe files underneath. Need an alias on the recipes object to a recipe file for the .default() recipe

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,111 @@
...
+```
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+stub = stub + "?" + "checkpoint=preqat"
+# will read all the files from the server, except for the full training folder. Training folder
+# will be identical to prequat checkpoint directory
+model = Model(path=stub)
+training: str = model.training.downloaded_path()
```
markurtz, line 26: let's do model.training.default() to match recipes and to have a clear mapping back to folder structure of what happened here

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,111 @@
...
+...
+```
+### Through the API purely
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+model = Model(path=stub)
+# ALTERNATIVE ONE: filtering on download
+model.recipes.download(recipe="transfer_learn")
```
markurtz, line 36: model.recipes["transfer_learn"].path

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,111 @@
...
+recipe: str = model.recipes.downloaded_path()
+# ALTERNATIVE TWO: filtering on .downloaded_path()
+recipe: str = model.recipes.downloaded_path(recipe="transfer_learn")
+```
+
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+model = Model(path=stub)
```
markurtz, line 46: let's make sure Model(stub) works as well, so do not require named argument to construct

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Download multiple chosen files (also in parallel)
+
+### Through the string arguments
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+# choose the files through the string arguments
+stub = stub + "?" + "recipe=transfer_learn" + "&" + "checkpoint=preqat" + "&" + "recipe=original"
```
markurtz, line 62: per URL standards, param names must be unique -- overwrite with the last if this is not the case

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,111 @@
...
+```
+### Through the API purely
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+model = Model(path=stub)
+# explicitly define which files to download
+model.recipes.download()
+model.training.download(checkpoint="preqat")
```
markurtz, line 76: model.training["preqat"].path

src/sparsezoo/v2/test.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Download all files, no "filtering"
+```python
+from sparsezoo import Model
+
+stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate"
+model = Model(path=stub)
+model.download()
+training: str = model.training.downloaded_path() 
+deployment: str = model.deployment.downloaded_path()
```
markurtz, line 90: if deployment arg (future support) given, should support model.deployment.default.path

src/sparsezoo/v2/test.md
dbogunowicz: Coming to think about it, I'd rather make it:
`model.directory.get_path()` rather than `model.directory.path`

This is purely because of the fact, that when calling the `getter` in the latter case, sometimes we would need to go through the `download()` function. However, there is some internal logic in the codebase which is also calls `directory.path` on a few occasions. This situation requires introducing an argument (should I download on the call or not), clearly a sign that this should be a method, not a getter.

dbogunowicz: @markurtz also, to have all the questions in one place:
What would be the use case of representing the recipe as a string --> str(recipe) ?

src/sparsezoo/v2/test.md
bfineran: renaming to `path_or_stub` should be fine, there shouldn't be a need to explicitly name this variable

dbogunowicz: yeah, the user can put anything in there and will not care about naming

src/sparsezoo/v2/test.md
dbogunowicz: (also model cards)

src/sparsezoo/v2/test.md
markurtz: model.download should download all files -- no files are private

src/sparsezoo/v2/test.md
markurtz: Let's keep it at just recipe, checkpoint, and deployment for now

PR #201, opened 2022-08-03
Make stub string key `recipe` go with values with more versatile names

dbogunowicz:
Currently:
`Model._validate_params()` will throw an error if `Model` receives any other recipe string arguments other than `?recipe=original` or `?recipe=transfer_learn`.

After the patch:
We account for the fact that we may be receiving arguments with more versatile names, such as `?recipe=transfer-text_classification` or `?recipe=original_fizz_buzz`

### DIFF
--- a/src/sparsezoo/model/model.py
+++ b/src/sparsezoo/model/model.py
@@ -41,7 +41,7 @@ from sparsezoo.validation import IntegrationValidator
 __all__ = ["Model"]
 
 ALLOWED_CHECKPOINT_VALUES = {"prepruning", "postpruning", "preqat", "postqat"}
-ALLOWED_RECIPE_VALUES = {"original", "transfer_learn"}
+ALLOWED_RECIPE_VALUES = {"original", "transfer"}
 ALLOWED_DEPLOYMENT_VALUES = {"default"}
 
 PARAM_DICT = {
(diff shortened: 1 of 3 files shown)


src/sparsezoo/model/model.py
```diff
@@ -700,9 +700,17 @@ def _validate_params(params: Dict[str, str]):
...
                     f"expected set of string arguments {list(PARAM_DICT.keys())}!"
                 )
-            if PARAM_DICT[key] and value not in PARAM_DICT[key]:
-                raise ValueError(
-                    f"String argument {key} has value {value}, "
-                    "cannot be found in the "
-                    f"expected set of values {list(PARAM_DICT[key])}!"
-                )
+            if key == "recipe":
+                if not any(identifier in value for identifier in PARAM_DICT[key]):
```
dbogunowicz, line 704: @bfineran do we have a convention for naming recipes? i.e
 every transfer learning recipe will be named `recipe_transfer_{...}` and every default recipe will be named `recipe_original_{...}`? If yes, we can change `identifier in value` to `value.startswith(identified)`.

bfineran: `recipe_` is not present in any of these names right now (might be in update). `transfer` should be present in general, but this is not a hard rule. I think this hard check is fine for now. better than what we had before. can relax further as needed when new recipes are added

bfineran (dismissed):

src/sparsezoo/model/model.py
markurtz: Standards moving forward will be renaming the base recipe to recipe.md and we will still need to support recipe_original for backwards compatibility. recipe_{name} is the future general format and recipe_transfer is a special match for transfer types that the user can request for use

markurtz: We should remove ALLOWED_RECIPE_VALUES and respect anything that is listed from the API. Big thing is making sure that when people request a transfer checkpoint that we don't require the full transfer name and can assume which one to return. If there are multiple there, then we fail

markurtz (changes requested):
Comments left in convo

bfineran (approved):
approving - confirming new logic is for stubs with a `recipe` param, we will return the recipe that matches `recipe_{recipe}`, if more than one match then we raise an exception

markurtz (approved):
Approved, need to do some regression testing for older recipe flows to make sure those are working with this

PR #220, opened 2022-08-17
Analysis tool: Handle invalid shapes, better erroring if ort is not installed

kylesayrs:
Previously the tool would error out after attempting to calculate the ops for a model if onnxruntime was not installed. Now, warnings make it clear that shape inference without ort is not preferred and specifies which nodes (if any) had to be skipped because of their invalid shapes.

```
WARNING:sparsezoo.utils.node_inference:Extracting shapes using ONNX Runtime session failed: No module named 'onnxruntime'
WARNING:sparsezoo.utils.node_inference:Falling back to ONNX shape_inference
WARNING:sparsezoo.utils.calculate_ops:Invalid shape, skipping ops calculation for Conv_0
```

I've tested by running analysis with one of every kind of architecture and none of the tested models produced an error.

### DIFF
--- a/src/sparsezoo/utils/calculate_ops.py
+++ b/src/sparsezoo/utils/calculate_ops.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import copy
+import logging
 from copy import deepcopy
 from typing import Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


bfineran (approved):
lgtm - please add brief pr description before land

KSGulin (approved):

dbogunowicz (approved):

PR #226, opened 2022-08-22
Make `sample_batch` method more intuitive

dbogunowicz:
From the user perspective, I felt that it is really annoying that the method:
```python
model.sample_batch(5)
```
does not do what I would suppose it naturally should do,  i.e. return a batch of data where `batch_size` is 5.

As of now, the first argument is the batch index, which I find super counterintuitive. 
To my mind, without reading the docstrings or looking at arguments' names, it feels natural that the first argument should be the batch size.

### DIFF
--- a/src/sparsezoo/model/model.py
+++ b/src/sparsezoo/model/model.py
@@ -372,7 +372,7 @@ class Model(Directory):
         )
 
     def sample_batch(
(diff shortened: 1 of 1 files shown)


bfineran (approved):
LGTM. Be sure to update any usages

corey-nm (approved):
Awesome change

PR #227, opened 2022-08-23
Restore test mode when SPARSEZOO_TEST_MODE is true

dbogunowicz:
Testing:
Together with @kazukageorge , we tested in staging/prod that:
- there are entries in the db when I run the test suite with `SPARSEZOO_TEST_MODE=false`
- there are no entries in the db when I run the test suite with `SPARSEZOO_TEST_MODE=true`

### DIFF
--- a/src/sparsezoo/utils/__init__.py
+++ b/src/sparsezoo/utils/__init__.py
@@ -17,12 +17,14 @@
 
 import os
 
+from sparsezoo.utils.helpers import convert_to_bool
 
 BASE_API_URL = (
     os.getenv("SPARSEZOO_API_URL")
     if os.getenv("SPARSEZOO_API_URL")
     else "https://api.neuralmagic.com"
 )
+SPARSEZOO_TEST_MODE = convert_to_bool(os.getenv("SPARSEZOO_TEST_MODE"))
 MODELS_API_URL = f"{BASE_API_URL}/models"
 LATEST_PACKAGE_VERSION_URL = f"{BASE_API_URL}/packages/check-latest"
 
--- a/src/sparsezoo/utils/__init__.py
+++ b/src/sparsezoo/utils/__init__.py
@@ -24,7 +24,6 @@ BASE_API_URL = (
     if os.getenv("SPARSEZOO_API_URL")
     else "https://api.neuralmagic.com"
 )
-SPARSEZOO_TEST_MODE = convert_to_bool(os.getenv("SPARSEZOO_TEST_MODE"))
 MODELS_API_URL = f"{BASE_API_URL}/models"
 LATEST_PACKAGE_VERSION_URL = f"{BASE_API_URL}/packages/check-latest"
 
--- a/src/sparsezoo/utils/helpers.py
+++ b/src/sparsezoo/utils/helpers.py
@@ -14,9 +14,28 @@
 
 import glob
 import os
+from typing import Any
 
 
-__all__ = ["create_dirs", "create_parent_dirs", "clean_path", "remove_tar_duplicates"]
+__all__ = [
+    "create_dirs",
+    "create_parent_dirs",
+    "clean_path",
+    "remove_tar_duplicates",
+    "convert_to_bool",
+]
+
+
+def convert_to_bool(val: Any):
+    """
+    :param val: a value
+    :return: False if value is a Falsy value e.g. 0, f, false, None, otherwise True.
+    """
+    return (
+        bool(val)
+        if not isinstance(val, str)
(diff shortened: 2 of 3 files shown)


src/sparsezoo/utils/requests.py
```diff
@@ -19,7 +19,7 @@
 
 import requests
 
-from . import MODELS_API_URL
+from . import MODELS_API_URL, SPARSEZOO_TEST_MODE
```
bfineran, line 22: nit: remove relative import (line below as well)

src/sparsezoo/utils/requests.py
```diff
@@ -106,15 +106,20 @@ def download_get_request(
...
         download_args.append(f"release_version={args.release_version}")
 
+    if SPARSEZOO_TEST_MODE:
+        # important, do not remove
+        download_args.append("increment_download=False")
+
     if download_args:
         url = f"{url}?{'&'.join(download_args)}"
 
+    print(url)
```
bfineran, line 116: debugging?

dbogunowicz:
There is still some uncertainty about whether tests do not "leak" when ran in GHA. I propose we land this PR and me with @kazukageorge look into GHA later.

kazukageorge:
So it passed on Damian's local testing machine (did not write to staging and prod db with tests), but it looks like github actions causing the write to the db. Hence, code should be fine, might be some set up (ex. env vars) issue on github actions' machine

src/sparsezoo/utils/requests.py
```diff
@@ -106,6 +106,10 @@ def download_get_request(
     if hasattr(args, "release_version") and args.release_version:
         download_args.append(f"release_version={args.release_version}")
 
+    if SPARSEZOO_TEST_MODE:
+        # important, do not remove
+        download_args.append("increment_download=False")
```
corey-nm, line 111: Another way we could do this in the future is by setting the User-Agent header to something specific to unit tests in this case, and then the server could check against that specific User-Agent. Then we wouldn't need a custom flag!

corey-nm (approved):

src/sparsezoo/utils/helpers.py
```diff
@@ -14,9 +14,28 @@
...
+__all__ = [
+    "create_dirs",
+    "create_parent_dirs",
+    "clean_path",
+    "remove_tar_duplicates",
+    "convert_to_bool",
+]
+
+
+def convert_to_bool(val: Any):
```
corey-nm, line 29: Maybe add some unit tests for this one? Seems like it should be pretty straightforward to test

src/sparsezoo/utils/__init__.py
```diff
@@ -17,12 +17,14 @@
...
 import os
 
+from sparsezoo.utils.helpers import convert_to_bool
 
 BASE_API_URL = (
     os.getenv("SPARSEZOO_API_URL")
     if os.getenv("SPARSEZOO_API_URL")
     else "https://api.neuralmagic.com"
 )
+SPARSEZOO_TEST_MODE = convert_to_bool(os.getenv("SPARSEZOO_TEST_MODE"))
```
corey-nm, line 27: Actually can we move this computation down into the actual function? That way the env variable is checked every time the function is called so we aren't dealing with some weird global variable caching

corey-nm:
> So it passed on Damian's local testing machine (did not write to staging and prod db with tests), but it looks like github actions causing the write to the db. Hence, code should be fine, might be some set up (ex. env vars) issue on github actions' machine

Which repo's github actions are you looking at? It seems like sparsezoo and sparseml both set the environment variable properly, but deepsparse does not. Could the updates be coming from deepsparse workflows?

kazukageorge:
> > So it passed on Damian's local testing machine (did not write to staging and prod db with tests), but it looks like github actions causing the write to the db. Hence, code should be fine, might be some set up (ex. env vars) issue on github actions' machine
> 
> Which repo's github actions are you looking at? It seems like sparsezoo and sparseml both set the environment variable properly, but deepsparse does not. Could the updates be coming from deepsparse workflows?

Thanks for the reply.

I re-ran the tests and it passed (did not write to the db). Something else was running independently causing the write to the db.

src/sparsezoo/utils/requests.py
bfineran: good idea, especially for outside testers /devs, but for now, still need ways to disable download increments for internal use + manual QA runs

corey-nm (approved):

bfineran (approved):

KSGulin (approved):

PR #232, opened 2022-08-25
Make `Model.eval_results` more robust

rahul-tuli:
In it's current form the codebase only relies on `eval.yaml` to get eval_results; this PR adds support to rely on. regex search for filenames containing eval/validation in them in case normal pathway fails

Also fixes a small typo in tests
Adds an extra test for eval results

### DIFF
--- a/src/sparsezoo/model/model.py
+++ b/src/sparsezoo/model/model.py
@@ -155,7 +155,7 @@ class Model(Directory):
         self.benchmarks: File = self._file_from_files(
             files, display_name="benchmarks.yaml"
         )
-        self.eval_results: File = self._file_from_files(files, display_name="eval.yaml")
+        self.eval_results: File = self._get_eval_results(files)
 
         # sorting name of `sample_inputs` and `sample_output` files,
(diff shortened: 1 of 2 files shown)


dbogunowicz:
Before we jump the gun, I would to understand why is there a `eval_results.json` file in the first place.
The source doc (https://docs.google.com/document/d/1kyuZ8EhF8L6CEnpCQ_zhrT4cu-PHwPJX7fJXHcM0dL0/edit#heading=h.gxfi15nnqdl6) as of today does not mention it, but the older version stated that the name for an evaluation run on models should save its results to `eval.json`. Before we land this, I will talk to Danny and Mark about the expected naming convention.

dbogunowicz:
@rahul-tuli where does the `eval_results.json` file come from?

rahul-tuli:
> @rahul-tuli where does the `eval_results.json` file come from?

It comes from the artifacts downloaded using the failing stub

rahul-tuli:
This Pr is not needed anymore

PR #234, opened 2022-08-26
Refactoring & testing utils/authentication.py

corey-nm:
This is PR is a refactor of the `sparsezoo/utils/authentication.py` module, as discussed in a previous PR. It also adds unit tests for all the functionality.

Summary:
- Replaced SparseZooCredentials class with two functions:
    - `_maybe_load_token` which tries to read in token from a path
    - `_save_token` which saves a token to a path
- The logic for loading a token is much more linear now, instead of spread across multiple methods

### DIFF
--- a/src/sparsezoo/utils/authentication.py
+++ b/src/sparsezoo/utils/authentication.py
@@ -45,77 +45,8 @@ CREDENTIALS_YAML = os.path.abspath(
 )
 
 
-class SparseZooCredentials:
-    """
-    Class wrapping around the sparse zoo credentials file.
-    """
-
-    def __init__(self):
-        if os.path.exists(CREDENTIALS_YAML):
-            _LOGGER.debug(f"Loading sparse zoo credentials from {CREDENTIALS_YAML}")
-            with open(CREDENTIALS_YAML) as credentials_file:
-                credentials_yaml = yaml.safe_load(credentials_file)
-                if credentials_yaml and CREDENTIALS_YAML_TOKEN_KEY in credentials_yaml:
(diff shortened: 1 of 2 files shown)


src/sparsezoo/utils/authentication.py
```diff
@@ -124,24 +55,63 @@ def get_auth_header(
 
     Currently only 'public' authentication type is supported.
 
-    :param authentication_type: authentication type for generating token
```
corey-nm, line 127: I removed authentication_type as a parameter since there was only 1 valid type and nowhere uses the parameter. Seem okay?

rahul-tuli (approved):

src/sparsezoo/utils/authentication.py
```diff
@@ -124,24 +55,63 @@ def get_auth_header(
...
+
+    info = creds[CREDENTIALS_YAML_TOKEN_KEY]
+    if "token" not in info or "created" not in info:
+        _LOGGER.debug(f"No sparse zoo credentials found at {path}")
+        return None
+
+    date_created = datetime.fromtimestamp(info["created"], tz=timezone.utc)
+    creation_difference = datetime.now(tz=timezone.utc) - date_created
+
+    if creation_difference.days > 30:
```
KSGulin, line 97: docstring for `get_auth_header()` cites that tokens older than 1 day will be updated. Code or doc should be updated

KSGulin (commented):
One nit + it looks like a quality check failed, but otherwise LGTM 🙂

KSGulin (approved):

jeanniefinks:
Product being deprecated by June 2, 2025. Closing PR and will not resolve.

PR #252, opened 2023-01-27
Performance analysis schema definitions

rahul-tuli:
This PR will add schema definitions for upcoming Performance Analysis api; These schemas will provide structure for different benchmarking experiments and the corresponding results.

Some major classes/schemas introduced in this work are:
-  `NodeInferenceResult`
-  `ImposedSparsificationInfo`
-  `BenchmarkSetup`
-  `BenchmarkResult`

We also add a `benchmark_results` field in existing `ModelAnalysis` class that will be populated later using a helper function.

Yaml serialization tests were also added across schemas

---
  - https://app.asana.com/0/0/1203803472587058


src/sparsezoo/analysis/analysis.py
```diff
@@ -54,12 +58,166 @@
...
+
+    num_streams: Optional[int] = Field(
+        default=None, description="Number of streams to use for benchmarking"
+    )
+
+    duration: int = Field(
+        default=10,
+        description="Number of seconds/steps the benchmark should run for, will use "
+        "steps instead of seconds if `duration_in_steps` is `True`; "
+        "defaults to 10 secs or steps",
```
KSGulin, line 170: What does `secs or steps` mean here?

KSGulin (commented):
LGTM, approved pending comment

src/sparsezoo/analysis/analysis.py
rahul-tuli: `seconds or steps`, but should probably just say defaults to 10; will update thanks!

src/sparsezoo/analysis/analysis.py
```diff
@@ -54,12 +58,166 @@
...
+    batch_size: PositiveInt = Field(
+        default=1,
+        description="The batch size to use for benchmarking, defaults to 1",
+    )
+
+    num_cores: Optional[int] = Field(
+        default=None,
+        description="The number of cores to use for benchmarking, defaults "
+        "to `None`, which represents all cores",
+    )
```
bfineran, line 148: let's not default batch size or num cores

bfineran (commented):
LGTM pending comment and a test for `BenchmarkResult`

bfineran (approved):

KSGulin (approved):

PR #256, opened 2023-02-08
[standardization-utils] markdown table util function

bfineran:
helper function for generating a markdown table from a list of headers and rows

**example_usage:**
python:
```python
from sparsezoo.utils.standardization import create_markdown_table

table = create_markdown_table(["sample", "table", "headers"], [["sample", "table", "row"]] * 3)
print(table)
```
rendered:
| sample | table | headers |
| ------ | ----- | ------- |
| sample | table | row     |
| sample | table | row     |
| sample | table | row     |

**test_plan:**
feature branch testing to be covered via integration test


KSGulin (approved):
Looks good!

PR #244, opened 2022-12-07
graphql requests module

kazukageorge:
Use v2 api endpoints for search and downloads. 

Please checkout the graphiql interactive UI to play around with the apis. 
https://staging-api.neuralmagic.com/v2/graphql

Example query:
```
    {
        models (domain: "cv",task: "classification",architecture: "resnet_v1",subArchitecture: "50",framework: "pytorch",repo: "sparseml",sourceDataset: "imagenet",sparseTag: "pruned95_uniform_quant-none",)
            {
               modelId modelOnnxSizeCompressedBytes stub 
            }
    }
```
Screen Shot 2023-02-03 at 10 19 12 AM

PR #270, opened 2023-02-14
Graphql queryparser fields

kazukageorge:
Blocked on https://github.com/neuralmagic/sparsezoo/pull/244

Allow the query parser to accept both list and dict for the fields


tdg5:
Heads up, I changed this PR to target the branch from #244 to make it easier to review the relevant differences.

src/sparsezoo/api/query_parser.py
```diff
@@ -93,12 +93,64 @@ def _parse_arguments(self) -> None:
...
+            fields = {
+                "model_id": None,
+                "benchmarks": {
+                    benchmark_result_id: None,
+                    whatever: {
+                        ...
+                    }
+                }
+            }
+            """
```
tdg5, line 109: What you have here is fine, don't feel like you need to change it, but I wanted to take the opportunity to point out an alternative as some food for thought. Some kinds of `if` statements can be replaced with a dictionary lookup. I think something like that could be possible here.
```python
FIELDS_PARSERS = {
    List: parse_list_fields_to_string,
    Dict: parse_dict_fields_to_string,
}
...

for fields_type, parser in FIELDS_PARSERS.items():
    if isinstance(fields, fields_type):
        self.fields = parser(fields)
        break
```
One benefit is that it makes the `_parse_fields` method better support the open/closed principle in that the addition of `FIELDS_PARSERS` makes it so new fields formats could be supported without having to change the logic in this method.

I defer to you on whether to take a similar approach. It's a bit awkward here since the parser functions are instance methods, if they could be made static methods or independent functions, that would help, but might come with its own set of tradeoffs.

src/sparsezoo/api/query_parser.py
```diff
@@ -93,12 +93,64 @@ def _parse_arguments(self) -> None:
 
     def _parse_fields(self) -> None:
```
tdg5, line 94: YMMV, but you might consider moving the `self.fields = ...` side effect of this function up to the caller or updating the method name to give a clue of the side effect. Maybe something like `_parse_and_set_fields` or something. I think I usually try to move the assignment up to the caller because I struggle to find a name for a method that does something and then makes an assignment. It can often be good to limit the number of functions that make assignments to make the lifecycle of the object easier to reason about.

src/sparsezoo/api/query_parser.py
```diff
@@ -93,12 +93,64 @@ def _parse_arguments(self) -> None:
...
             parsed_fields += f"{camel_case_field} "
-        self.fields = parsed_fields
+            if camel_case_field in DEFAULT_FIELDS:
+                parsed_fields += (
+                    "{ "
+                    + self.parse_list_fields_to_string(
+                        DEFAULT_FIELDS.get(camel_case_field)
+                    )
+                    + "} "
+                )
```
tdg5, line 124: I think I would pull out a local variable here:
```suggestion
                fields_for_field = self.parse_list_fields_to_string(
                    DEFAULT_FIELDS.get(camel_case_field)
                )
                parsed_fields += f"{{ {fields_for_field}}} "
```

I tried to match your spacing around the braces. Is that intentional spacing?

src/sparsezoo/api/query_parser.py
```diff
@@ -93,12 +93,64 @@ def _parse_arguments(self) -> None:
...
+        for field, field_dict in fields.items():
+            if field_dict:
+                assert isinstance(field_dict, Dict), (
+                    "QueryParser fields must be either dict where ",
+                    "values are a dict or None",
+                )
+
+                parent_field = to_camel_case(field)
+                child_fields = (
+                    " { " + self.parse_dict_fields_to_string(field_dict) + "} "
```
tdg5, line 138: The spacing around braces here is different than the spacing around braces above. I don't know if it matters, but wanted to bring the difference to your attention.

Again, I might introduce a local variable here, but doesn't strike me as critical.

src/sparsezoo/api/query_parser.py
```diff
@@ -93,12 +93,64 @@ def _parse_arguments(self) -> None:
...
+                camel_case_field = to_camel_case(field)
+                parsed_fields += f"{camel_case_field} "
+                if camel_case_field in DEFAULT_FIELDS:
+                    parsed_fields += (
+                        "{ "
+                        + self.parse_list_fields_to_string(
+                            DEFAULT_FIELDS.get(camel_case_field)
+                        )
+                        + "} "
+                    )
```
tdg5, line 151: This is the same logic from above, no? Maybe a helper method would help DRY things? Not sure if there's value in that, but worth considering in this case I'd think.

tdg5 (approved):

tdg5 (approved):

PR #271, opened 2023-02-16
Stub v2

kazukageorge:
Make stub v2 compatible


tdg5:
Heads up, I changed this PR to target the branch from #270 to make it easier to review the relevant differences.

src/sparsezoo/api/utils.py
```diff
@@ -34,5 +34,20 @@ def to_snake_case(string: str):
...
+    Given a dictionary, update its keys to a given mapper callable.
+        If the value of the dict is a List of Dict or Dict of Dict,
+        recursively map its keys
+    """
+    mapped_dict = {}
+    for key, value in dictionary.items():
+        if isinstance(value, List) or isinstance(value, Dict):
+            value_type = type(value)
+            mapped_dict[mapper(key)] = value_type(
+                map_keys(dictionary=sub_dict, mapper=to_snake_case)
```
tdg5, line 47: I would expect `mapper=mapper` here:
```suggestion
                map_keys(dictionary=sub_dict, mapper=mapper)
```

tdg5: I don't know that there's necessarily something to do about it... I guess you could add another `if` statement or something? but if type checking were being enforced, this call would be a type error. `field` is a `List[Any]` as far as I can tell, and yet `map_keys` is called as though all of the children of the list are of type `Dict`. I guess the type annotation for `dictionary` isn't right also? Doesn't seem like `Dict[str, str]` leaves any room for `Dict` or `List` to show up as fields.

src/sparsezoo/model/utils.py
```diff
@@ -120,31 +139,29 @@ def load_files_from_stub(
...
+        fields=[
+            "model_id",
+            "model_onnx_size_compressed_bytes",
+            "files",
+            "benchmark_results",
+            "training_results",
+        ],
     )
 
     if len(models):
```
tdg5, line 151: Am I grumpy old man for wanting to raise an error if there are many matching models instead of just rolling with the first model in the list? I don't know what the right answer is.

src/sparsezoo/model/utils.py
```diff
@@ -120,31 +139,29 @@ def load_files_from_stub(
...
+
         model_id = models[0]["model_id"]
```
tdg5, line 153: I'd recommend pulling out a local variable instead of repeatedly invoking `models[0]`
```python
model = models[0]
```

src/sparsezoo/model/utils.py
```diff
@@ -562,29 +579,31 @@ def include_file_download_url(files: List[Dict]):
...
+                "sub_domain": matches.group("sub_domain"),
+                "architecture": matches.group("architecture"),
+                "sub_architecture": matches.group("sub_architecture"),
+                "framework": matches.group("framework"),
+                "repo": matches.group("repo"),
+                "dataset": matches.group("dataset"),
+                "sparse_tag": matches.group("sparse_tag"),
+            }
+
+    return {}
```
tdg5, line 601: I'm tempted to untangle the handling of the two different formats. I might propose handling the new format separately from the old format:
```suggestion
    if re.match(STUB_V2_REGEX_EXPR, stub):
        return {"repo_name": stub}
    
    matches = re.match(STUB_V1_REGEX_EXPR, stub)
    if matches:
        return {
            "domain": matches.group("domain"),
            "sub_domain": matches.group("sub_domain"),
            "architecture": matches.group("architecture"),
            "sub_architecture": matches.group("sub_architecture"),
            "framework": matches.group("framework"),
            "repo": matches.group("repo"),
            "dataset": matches.group("dataset"),
            "sparse_tag": matches.group("sparse_tag"),
        }

    return {}
```

If you were to take this approach, you might not need any capture groups in `STUB_V2_REGEX_EXPR`

tdg5: Also, I don't know if it's important, but I usually think of `training_scheme` as a stub component. Doesn't seem like it really comes up in the data though.

src/sparsezoo/model/utils.py
```diff
@@ -562,29 +579,31 @@ def include_file_download_url(files: List[Dict]):
...
+
+
+def is_stub(candidate: str):
```
tdg5, line 604:
```suggestion
def is_stub(candidate: str) -> bool:
```

src/sparsezoo/model/utils.py
```diff
@@ -562,29 +579,31 @@ def include_file_download_url(files: List[Dict]):
...
+    if candidate.startswith(ZOO_STUB_PREFIX):
+        return True
+    if bool(get_model_metadata_from_stub(candidate)):
+        return True
+    return False
```
tdg5, line 609: I don't know if it matters, but I think I'd handle this `if` statement by directly checking the string instead of seeing if a full metadata dict can be extracted from candidate
```suggestion
def is_stub(candidate: str) -> bool:
    return re.match(STUB_V1_REGEX_EXPR, candidate) or re.match(STUB_V2_REGEX_EXPR, candidate)
```

tests/sparsezoo/model/test_model.py
```diff
@@ -141,6 +151,21 @@ def _assert_validation_results_exist(model):
             True,
             files_yolo,
         ),
+        (
+            "mobilebert-squad_wikipedia_bookcorpus-14layer_pruned50.4block",  # noqa E501
+            False,
+            files_yolo,
+        ),
+        (
+            "yolov5-s-coco-pruned85_quantized",  # noqa E501
```
tdg5, line 160: Is the `noqa` really needed here?

src/sparsezoo/model/utils.py
tdg5: Re: `training_scheme`, not sure if there's a test for extracting metadata... but you might consider a test case that ensures that `zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet-augmented/pruned_quant-aggressive` is handled correctly and yields distinct metadata from `zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned_quant-aggressive`

tdg5 (changes requested):
I think there are a couple of bugs in here, but overall makes sense to me

src/sparsezoo/model/utils.py
```diff
@@ -118,33 +138,45 @@ def load_files_from_stub(
...
 
-        benchmark_results = api.fetch(
-            operation_body="benchmark_results",
-            arguments={"model_id": model_id},
-        )
+        benchmark_results = model.get("benchmark_results")
 
-        model_onnx_size_compressed_bytes = models[0]["model_onnx_size_compressed_bytes"]
+        model_onnx_size_compressed_bytes = model.get("model_onnx_size_compressed_bytes")
+        print(model_onnx_size_compressed_bytes)
```
tdg5, line 179:
```suggestion
```

src/sparsezoo/model/utils.py
```diff
@@ -553,6 +585,36 @@ def include_file_download_url(files: List[Dict]):
...
+                "sub_domain": matches.group("sub_domain"),
+                "architecture": matches.group("architecture"),
+                "sub_architecture": matches.group("sub_architecture"),
+                "framework": matches.group("framework"),
+                "repo": matches.group("repo"),
+                "dataset": matches.group("dataset"),
+                "sparse_tag": matches.group("sparse_tag"),
+            }
+
+    return {}
```
tdg5, line 607: You can reduce nesting if you short-circuit this condition:
```suggestion
    if not matches:
        return {}

    if "source_dataset" in matches.groupdict():
        return {"repo_name": stub}

    if "dataset" in matches.groupdict():
        return {
            "domain": matches.group("domain"),
            "sub_domain": matches.group("sub_domain"),
            "architecture": matches.group("architecture"),
            "sub_architecture": matches.group("sub_architecture"),
            "framework": matches.group("framework"),
            "repo": matches.group("repo"),
            "dataset": matches.group("dataset"),
            "sparse_tag": matches.group("sparse_tag"),
        }

    return {}
```

Maybe the last `return {}` should be a `raise` instead?

kazukageorge: returning no metadata should be appropriate

src/sparsezoo/api/utils.py
```diff
@@ -32,7 +32,21 @@ def to_snake_case(string: str):
...
 def map_keys(
-    dictionary: Dict[str, str], mapper: Callable[[str], str]
+    dictionary: Dict[str, Any], mapper: Callable[[str], str]
 ) -> Dict[str, str]:
-    """Given a dictionary, update its key to a given mapper callable"""
-    return {mapper(key): value for key, value in dictionary.items()}
+    """
+    Given a dictionary, update its keys to a given mapper callable.
+        If the value of the dict is a List of Dict or Dict of Dict,
+        recursively map its keys
```
tdg5, line 40: Is this indentation typical?
```suggestion
    Given a dictionary, update its keys to a given mapper callable.

    If the value of the dict is a List of Dict or Dict of Dict, recursively map
    its keys
```

tdg5 (approved):

dfneuralmagic (approved):
LGTM

PR #283, opened 2023-03-22
fix type

kazukageorge:


derekk-nm:
It would be useful to have some tests here that verify this change is working correctly.

src/sparsezoo/api/graphql.py
```diff
@@ -28,7 +28,7 @@ def fetch(
         self,
         operation_body: str,
         arguments: Optional[Dict[str, str]] = None,
-        fields: Optional[List[str]] = None,
+        fields: Optional[Union[List[str], Dict[str, Union[Dict, None]]]] = None,
```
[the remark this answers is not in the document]
kazukageorge: graphql will not handle either, but we have a parser that given Dict or List, will convert the input fields to a graphql compatible format.

kazukageorge:
> It would be useful to have some tests here that verify this change is working correctly.

There already are tests to check both List and Dict. 
https://github.com/neuralmagic/sparsezoo/blob/main/tests/sparsezoo/api/test_query_parser.py#L122-L126

tdg5 (approved):

src/sparsezoo/api/graphql.py
```diff
@@ -72,6 +72,9 @@ def make_request(
             operation_body=operation_body, arguments=arguments, fields=fields
         )
 
+        print(query.arguments)
```
derekk-nm, line 75: These lines look like you were debugging?

derekk-nm (dismissed):

src/sparsezoo/api/graphql.py
```diff
@@ -27,8 +27,8 @@ class GraphQLAPI:
     def fetch(
         self,
         operation_body: str,
-        arguments: Optional[Dict[str, str]] = None,
-        fields: Optional[List[str]] = None,
+        arguments: Optional[Dict[str, Any]] = None,
+        fields: Optional[Union[List[str], Dict[str, Union[Dict, None]]]] = None,
```
tdg5, line 31: Can you get away with `Optional[Dict]` and avoid the need for one of the `Union`?
```suggestion
        fields: Optional[Union[List[str], Dict[str, Optional[Dict]]]] = None,
```

tdg5 (dismissed):

derekk-nm (approved):

PR #293, opened 2023-04-05
[analytics] disable in make test and GHA tests, unit test to always check send_event

bfineran:
adds env var to disable analytics calls in GHA tests and `make test`

adds unit test to override the disable and verify a simple `send_event` call. add pathway to ensure that analytics thread completes

**test_plan:**
* unit tests included
* @markurtz to verify GHA no longer triggers GA calls


mgoin (approved):

rahul-tuli (approved):
Nice changes

PR #294, opened 2023-04-07
add device_info to _extract_throughput

bfineran:
sparsezoo backend update now pushes more throughput results for each model, updates `_extract_throughput` to account for device info and fixes broken test

**test_plan:**
unit test updated


bfineran:
tests green - merging

rahul-tuli (commented):
Nice Change!

PR #308, opened 2023-04-26
Helper functions for large ONNX models support

dbogunowicz:
Implementing onnx helper functions that will be shared across our repositories (deepsparse, sparseml, etc.) to reliably save and validate not only the models which fit into a protobuf (as we did up until now) but also the larger ones (where the large data information is stored in the additional accompanying file).

Did not include the helper function for loading onnx models, since this is not needed if we work under the assumption  that the onnx file and the auxiliary data file are in the same directory, see: https://github.com/onnx/onnx/blob/main/docs/ExternalData.md

Related PRs: 
https://github.com/neuralmagic/deepsparse/pull/995 (deepsparse)
https://github.com/neuralmagic/sparseml/pull/1540 (sparseml)

Manual testing (only the big model pathway):
```python
import shutil
from sparsezoo.utils.onnx import save_onnx, validate_onnx
import os
import onnx
import glob

### Validate the new functions for >2Gb models ###

## Prepare the model and test directory ##
model_path = "/network/damian/big_models/codegen-2B-multi/decoder_model.onnx"
model_onnx = onnx.load(model_path)
test_dir = "test_dir"

## Test validate_onnx ##
# should pass for a big model, input is path
validate_onnx(model = model_path)
print("Test 1 passed")

# should raise warning for a big model, input is model
validate_onnx(model = model_onnx)
print("Test 2 passed")

## Test save_onnx ##
# should produce two files
os.mkdir(test_dir)
save_onnx(model_onnx, os.path.join(test_dir, "test.onnx"))
files = glob.glob(os.path.join(test_dir, "*"))
print(f"Produced files {files}")
shutil.rmtree(test_dir)
print("Test 3 passed")

#should produce two files as well
os.mkdir(test_dir)
save_onnx(model_onnx, os.path.join(test_dir, "test.onnx"), external_data_file="something")
files = glob.glob(os.path.join(test_dir, "*"))
print(f"Produced files {files}")
shutil.rmtree(test_dir)
print("Test 4 passed")
```
Out:
```bash
Test 1 passed
Test 2 passed
Attempting to validate an in-memory ONNX model with size > 2000000000 bytes.`validate_onnx` skipped, as large ONNX models cannot be validated in-memory. To validate this model, save it to disk and call `validate_onnx` on the file path.
Produced files ['test_dir/test.onnx', 'test_dir/963eac78-e441-11ed-8c90-d52fb1102de6']
Test 3 passed
Produced files ['test_dir/test.onnx', 'test_dir/something']
Test 4 passed
```


src/sparsezoo/utils/onnx.py
```diff
@@ -42,6 +50,100 @@
...
+    model: ModelProto, model_path: str, external_data_file: Optional[str] = None
+) -> bool:
+    """
+    Save model to the given path.
+
+    If the model's size is larger than the maximum protobuf size:
+        -   it will be saved with external data
+    If the model's size is smaller than the maximum protobuf size:
+        -   and the user nevertheless specifies 'external_data_file',
+            the model will be saved with external data.
```
dbogunowicz, line 63:
```suggestion
            the model will be saved with external data.
    
```

src/sparsezoo/utils/onnx.py
```diff
@@ -42,6 +50,100 @@
...
+        )
+        return True
+
+    if model.ByteSize() > onnx.checker.MAXIMUM_PROTOBUF:
+        _LOGGER.info(
+            "The ONNX model is too large to be saved as a single protobuf."
+            "Saving with external data... "
+        )
+
+        _LOGGER.info(f"Saving with external data: {external_data_file}")
```
KSGulin, line 87: Wouldn't `external_data_file` evaluate to `None` here?

src/sparsezoo/utils/onnx.py
```diff
@@ -42,6 +50,100 @@
...
+    :param model: the model proto or path to the model
+        ONNX file to check for validation
+    """
+    try:
+        onnx_model = load_model(model)
+        if onnx_model.ByteSize() > onnx.checker.MAXIMUM_PROTOBUF:
+            if isinstance(model, str):
+                onnx.checker.check_model(model)
+            else:
+                _LOGGER.warning(
```
KSGulin, line 114: Nice warning

src/sparsezoo/utils/onnx.py
```diff
@@ -42,6 +50,100 @@
...
+                    "Attempting to validate an in-memory ONNX model with "
+                    f"size > {onnx.checker.MAXIMUM_PROTOBUF} bytes."
+                    "`validate_onnx` skipped, as large ONNX models cannot "
+                    "be validated in-memory. To validate this model, save "
+                    "it to disk and call `validate_onnx` on the file path."
+                )
+            return
+        onnx.checker.check_model(onnx_model)
+        if not onnx_model.opset_import:
+            raise ValueError("could not parse opset_import")
```
KSGulin, line 124: What is this guarding against?

KSGulin (commented):
Looks good! Left a couple comments

src/sparsezoo/utils/onnx.py
dbogunowicz: This was there before in the original function (https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/onnx/utils/helpers.py#L98), not really sure what's the function here.
I think Mark is the one who wrote it and added it in this commit: https://github.com/neuralmagic/sparseml/commit/4c1ae34616db21cb594668fb82c588b9a43d7c94

src/sparsezoo/utils/onnx.py
dbogunowicz: You are right, good catch. Will just remove this line of code.

dbogunowicz, line 87:
```suggestion
```

src/sparsezoo/utils/onnx.py
KSGulin: Ah ok, I guess it's just checking that the model has an opset number

KSGulin (dismissed):
Approved 🙂

bfineran (commented):
LGTM pending comments

src/sparsezoo/utils/onnx.py
```diff
@@ -42,6 +50,99 @@
...
+        -   and the user nevertheless specifies 'external_data_file',
+            the model will be saved with external data.
+
+    :param model: The model to save.
+    :param model_path: The path to save the model to.
+    :param external_data_file: The optional name save the external data to.
+    :return True if the model was saved with external data, False otherwise.
+    """
+    if external_data_file is not None:
+        _LOGGER.info(f"Saving with external data: {external_data_file}")
```
dbogunowicz, line 71:
```suggestion
        _LOGGER.debug(f"Saving with external data: {external_data_file}")
```

bfineran (approved):

PR #315, opened 2023-05-18
[bugfix] defer matplotlib import error until used

rahul-tuli:
Right now, invoking sparsezoo.analyze --help without dev dev dependencies fails as matplotlib won't be installed and it's needed to plot graphs, this PR defers the import error until matplotlib is actually used

---
  - https://app.asana.com/0/0/1204623202382276
  - https://app.asana.com/0/0/1204623198603898

### DIFF
--- a/src/sparsezoo/analyze/utils/chart.py
+++ b/src/sparsezoo/analyze/utils/chart.py
@@ -44,7 +44,7 @@ def check_matplotlib_installed() -> None:
     if not matplotlib_available:
         raise ImportError(
             "matplotlib is required to use this function, "
-            "please install it with `pip install matplotlib`"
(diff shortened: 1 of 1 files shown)


bfineran (approved):

KSGulin (approved):
LGTM

dbarbuzzi:
It would be nice to update the message to properly quote the module installed because the use of angle brackets in the command (e.g. `>`) without being quoted will not work. This will be greatly helpful to anyone that copies the install command from the error message to then perform the install.

For example:

```bash
# This does not use the constraint correctly; at best it installs with no version constraint and pipes all stdout to a file named `=3.0.0`
pip install matplotlib>=3.0.0
# This is the correct command for any standard shell
pip install "matplotlib>=3.0.0"
```

PR #317, opened 2023-05-24
Make 'validate_onnx' function robust against the models loaded without external data

dbogunowicz:
Accounting for the scenario, where the in-memory model passed to the `validate_onnx` is loaded without the external data.
Right now, the validator simply fails. What it should do, it should skip the validation and emit a verbose warning - this is what has been implemented.

Testing:
```python
import onnx
from sparsezoo.utils.onnx import validate_onnx
model = onnx.load("/network/damian/sparsegpt_webinar_fp32/sparsegpt_1.3b/model.onnx", load_external_data=False)
validate_onnx(model)
model_ = onnx.load("/network/damian/sparsegpt_webinar_fp32/sparsegpt_1.3b/model.onnx", load_external_data=True)
validate_onnx(model_)
```

```bash
Attempting to validate an in-memory ONNX model that has been loaded without external data. This is currently not supported by the ONNX checker. The validation will be skipped.
Attempting to validate an in-memory ONNX model with size > 2000000000 bytes.`validate_onnx` skipped, as large ONNX models cannot be validated in-memory. To validate this model, save it to disk and call `validate_onnx` on the file path.
```

### DIFF
--- a/src/sparsezoo/utils/onnx.py
+++ b/src/sparsezoo/utils/onnx.py
@@ -27,6 +27,7 @@ from sparsezoo.utils.helpers import clean_path
 _LOGGER = logging.getLogger(__name__)
 
 __all__ = [
+    "onnx_includes_external_data",
     "save_onnx",
     "validate_onnx",
     "load_model",
@@ -48,6 +49,30 @@ __all__ = [
     "extract_node_id",
     "get_node_attributes",
 ]
+
+
+def onnx_includes_external_data(model: ModelProto) -> bool:
+    """
+    Check whether the ModelProto in memory includes the external
+    data or not.
+
+    If the model.onnx does not contain the external data, then the
+    initializers of the model are pointing to the external data file
+    (they are not empty)
+
+    :param model: the ModelProto to check
+    :return True if the model was loaded with external data, False otherwise.
+    """
+
+    initializers = model.graph.initializer
+
+    data_saved_to_disk = any(
+        [initializer.external_data for initializer in initializers]
+    )
+
+    data_included_in_model = not data_saved_to_disk
+
+    return data_included_in_model
 
 
 def save_onnx(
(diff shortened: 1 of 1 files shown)


src/sparsezoo/utils/onnx.py
```diff
@@ -50,6 +51,30 @@
...
+    (they are not empty)
+
+    :param model: the ModelProto to check
+    :return True if the model was loaded with external data, False otherwise.
+    """
+
+    initializers = model.graph.initializer
+
+    data_saved_to_disk = any(
+        [initializer.external_data for initializer in initializers]
```
rahul-tuli, line 70:
```suggestion
        initializer.external_data for initializer in initializers
```
nit: let's take advantage of generators here?

src/sparsezoo/utils/onnx.py
```diff
@@ -121,6 +146,14 @@ def validate_onnx(model: Union[str, ModelProto]):
             return
         onnx.checker.check_model(onnx_model)
     except Exception as err:
```
rahul-tuli, line 148: Thought: This seems like a very broad except clause, could we shrink it down to a fewer more specific exceptions maybe?

rahul-tuli (commented):
A few suggestions, but the diff looks real nice!

KSGulin (commented):
LGTM pending Rahul's comments

src/sparsezoo/utils/onnx.py
dbogunowicz: Maybe, but I do not think we should work on this now, may be an act of overoptimization.

KSGulin (approved):

src/sparsezoo/utils/onnx.py
rahul-tuli: Still not convinced, but giving a green tick

rahul-tuli (approved):

PR #331, opened 2023-06-08
Deprecate Py37

rahul-tuli:
Python 3.7 reaches it's EOL in ~2 weeks, this PR deprecates it's support

https://endoflife.date/python

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -14,7 +14,7 @@ A clear and concise description of what you expected to happen.
 **Environment**
 Include all relevant environment information:
 1. OS [e.g. Ubuntu 18.04]:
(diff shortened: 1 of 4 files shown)


bfineran (approved):

dsikka:
Question @rahul-tuli seems like we skipped the `extensive-python-tests` check - if we're changing Python version support, just wondering why this test was skipped?

rahul-tuli:
> Question @rahul-tuli seems like we skipped the `extensive-python-tests` check - if we're changing Python version support, just wondering why this test was skipped?

Hi, great question, it's due to this line: https://github.com/neuralmagic/sparsezoo/blob/6fcc9dec9f4b1e9ca026fba340b000b86e6d90fd/.github/workflows/test-check.yaml#L31 which disables extensive tests when the branch is not main or release

dsikka (approved):
LGTM

PR #332, opened 2023-06-12
Update numpy 1.21.6

rahul-tuli:
Updates support to latest numpy
Needed to support deepsparse numpy version upgrade from neuralmagic/deepsparse#1094

The deepsparse change was made to support a ragged sequence being created in question answering pipeline during `spans_extra_info` computation; landing this diff along with   https://github.com/neuralmagic/sparseml/pull/1623 also fixes a few google colab errors we were seeing before.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ print(f"loaded version {version} from src/sparsezoo/version.py")
 _PACKAGE_NAME = "sparsezoo" if is_release else "sparsezoo-nightly"
 
 _deps = [
-    "numpy>=1.0.0,<=1.21.6",
+    "numpy>=1.0.0",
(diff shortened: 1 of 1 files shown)


SmartAudioTools:
Hello!
I'm still not able to install deepsparse with numpy>=1.22 with pip install. 
The pipy version does not seem to include the modification

ERROR: Cannot install deepsparse and numpy>=1.22 because these package versions have conflicting dependencies.

The conflict is caused by:
The user requested numpy>=1.22
deepsparse 1.5.2 depends on numpy>=1.16.3
onnx 1.12.0 depends on numpy>=1.16.6
sparsezoo 1.5.2 depends on numpy<=1.21.6 and >=1.0.0

PR #355, opened 2023-08-11
default Model.onnx_model to target onnx.model.tar.gz

bfineran:
currently, sparsezoo does not have great support for onnx models saved with external data out of the box. especially for user friendly and deepsparse compatible flows such as `Model.onnx_model.path`.

Better support will be added in a future backend update - to fix this now, this PR updates the default target for `Model.onnx_model` to be the `model.onnx.tar.gz` where possible.  A helper class is added to enforce that when a user tries to access the onnx model path, the compressed model is downloaded and unzipped and the reference to the model path points only to the extracted `model.onnx` file where applicable

**note - any users who have already tried to download models with external data will need to clear them from cache**

**test_plan:**
Manually verified with both an external model and a legacy model (see below) - would be good to add these as unit tests, however the download times for the external models can take a while as they are >2Gb. Existing functionality should be thoroughly covered in sparsezoo, deepsparse, and sparseml testing

```python
from sparsezoo import Model
import onnx

large_model_zoo = Model("zoo:nlg/text_generation/opt-1.3b/pytorch/huggingface/opt_pretrain/quantW8A8-none")
small_model_zoo = Model("zoo:cv/classification/vgg-19/pytorch/sparseml/imagenet/pruned-moderate")

large_model_onnx = onnx.load(large_model_zoo.onnx_model.path)
small_model_onnx = onnx.load(small_model_zoo.onnx_model.path)
```

### DIFF
--- a/src/sparsezoo/model/model.py
+++ b/src/sparsezoo/model/model.py
@@ -34,6 +34,7 @@ from sparsezoo.objects import (
     Directory,
     File,
     NumpyDirectory,
+    OnnxGz,
     SelectDirectory,
     is_directory,
 )
@@ -156,7 +157,14 @@ class Model(Directory):
             stub_params=self.stub_params,
         )
 
-        self.onnx_model: File = self._file_from_files(files, display_name="model.onnx")
+        self._onnx_gz: Directory = self._directory_from_files(
+            files, directory_class=OnnxGz, display_name="model.onnx.tar.gz"
+        )
+        self.onnx_model: File = (
+            self._file_from_files(files, display_name="model.onnx.")
+            if self._onnx_gz is None
+            else self._onnx_gz  # if onnx.model.tar.gz present defer to that file
+        )
 
         self.analysis: File = self._file_from_files(files, display_name="analysis.yaml")
         self.benchmarks: File = self._file_from_files(
--- a/src/sparsezoo/model/model.py
+++ b/src/sparsezoo/model/model.py
@@ -161,7 +161,7 @@ class Model(Directory):
             files, directory_class=OnnxGz, display_name="model.onnx.tar.gz"
         )
         self.onnx_model: File = (
-            self._file_from_files(files, display_name="model.onnx.")
+            self._file_from_files(files, display_name="model.onnx")
             if self._onnx_gz is None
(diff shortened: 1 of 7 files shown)


rahul-tuli (approved):

src/sparsezoo/objects/directories.py
```diff
@@ -280,16 +280,17 @@ class OnnxGz(Directory):
     """
     Special class to handle onnx.model.tar.gz files.
     Desired behavior is that all information about files included in the tarball are
-    available however, when the file.path is accessed, it will point only to the
-    `model.onnx` as this is the expected behavior for loading, additionally,
+    available however, when the file's `path` property is accessed, it will point only
```
bfineran, line 283: annoying, but apostrophes `'` are not well supported by Sphinx, they tend to parse these as unclosed strings

src/sparsezoo/model/model.py
```diff
@@ -156,7 +157,14 @@ def __init__(self, source: str, download_path: Optional[str] = None):
             stub_params=self.stub_params,
         )
 
-        self.onnx_model: File = self._file_from_files(files, display_name="model.onnx")
+        self._onnx_gz: Directory = self._directory_from_files(
+            files, directory_class=OnnxGz, display_name="model.onnx.tar.gz"
+        )
+        self.onnx_model: File = (
+            self._file_from_files(files, display_name="model.onnx.")
```
Satrat, line 164: should this be "model.onnx" instead of "model.onnx." ?

bfineran: good catch

bfineran, line 164:
```suggestion
            self._file_from_files(files, display_name="model.onnx")
```

src/sparsezoo/objects/directories.py
```diff
@@ -269,3 +274,23 @@ def available(self):
     @available.setter
     def available(self, value):
         self._available = value
+
+
+class OnnxGz(Directory):
+    """
+    Special class to handle onnx.model.tar.gz files.
+    Desired behavior is that all information about files included in the tarball are
+    available however, when the file's `path` property is accessed, it will point only
```
rahul-tuli, line 283:
```suggestion
    available however, when the `path` property is accessed, it will point only
```

PR #356, opened 2023-08-16
Remove dead code

tdg5:
I'm not sure if `sparsezoo.package` is provided so other libraries can import it, but if not, this code seems dead. Github search only yielded [one instance where I could see it was imported](https://github.com/neuralmagic/arm-competitive-benchmarking/blob/6d09a98ab60d81dfa9adff70d126726b2c1785cd/deepsparse/src/deepsparse/__init__.py#L40) and otherwise the other references seemed to be just `docs`

The backend API that I think `sparsezoo.package` uses is pretty defunct and I'd think this code isn't doing the right thing anyway.


jeanniefinks:
Product being deprecated by June 2, 2025. Closing PR and will not resolve.

PR #357, opened 2023-08-17
[Bugfix] Point OnnxGz files to `model.onnx` but download and extract `model.onnx.tar.gz`

rahul-tuli:
Big LLMS are generally saved with external data; this external data along with model.onnx live inside `model.onnx.tar.gz` tarball on sparsezoo; recently we added support to download the tarfile and emulate a normal onnx model file using a wrapper class `OnnxGz`; when adding this support a bug was introduced where `model.onnx` was present twice in the model path; leading to errors while creating ort sessions as 

### TEST PLAN:

- [X] Manual inspection using attached script
- [X] Add automated tests to capture conditions where extraction was failing
- [X] Test with deepsparse main

Test Script:
```
from sparsezoo import Model
from pathlib import Path


def local_test(stub: str):
    model = Model(stub)
    onnx_model = model.onnx_model
    model_path = onnx_model.path
    onnx_model_path = Path(model_path)
    assert onnx_model_path.exists(), "Path to onnx model does not exist"
    assert (count := str(onnx_model_path).count("model.onnx")) == 1, f"Path to onnx model contains model.onnx {count} times"

    # create an ORT session with the model
    import onnxruntime as ort
    session = ort.InferenceSession(str(onnx_model_path), providers=["CPUExecutionProvider"])
    assert session is not None, "ORT session is None"

if __name__ == '__main__':
    # stub = "zoo:nlg/text_generation/opt-13b/pytorch/huggingface/opt_pretrain/pruned50_quantW8A8-none"
    stub = "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned_quant-moderate"
    local_test(stub=stub)
```

Before this PR: The onnx_model_path had an extra model.onnx in it's path leading to errors like:
https://github.com/neuralmagic/deepsparse/actions/runs/5869604261/job/15914935604?pr=1150#step:7:138

After this PR the test script runs successfully, and the base test errors on deepsparse side are green

---
  - https://app.asana.com/0/0/1205276886236935

### DIFF
--- a/src/sparsezoo/model/model.py
+++ b/src/sparsezoo/model/model.py
@@ -157,7 +157,7 @@ class Model(Directory):
             stub_params=self.stub_params,
         )
 
-        self._onnx_gz: Directory = self._directory_from_files(
+        self._onnx_gz: OnnxGz = self._directory_from_files(
             files, directory_class=OnnxGz, display_name="model.onnx.tar.gz"
         )
         self.onnx_model: File = (
@@ -452,7 +452,7 @@ class Model(Directory):
     def _get_directory(
         self,
         file: Dict[str, Any],
-        directory_class: Directory,
+        directory_class: type(Directory),
         display_name: Optional[str] = None,
         regex: Optional[bool] = False,
         **kwargs,
@@ -587,12 +587,12 @@ class Model(Directory):
         regex: Optional[bool] = False,
         allow_multiple_outputs: Optional[bool] = False,
         **kwargs: object,
-    ) -> Union[Directory, None]:
+    ) -> Union[List[Union[Directory, Any, None]], List[Directory], None]:
 
         # Takes a list of file dictionaries and returns
         # a Directory() object, if successful,
         # otherwise None.
-        if all([file_dict.get("file_type") for file_dict in files]):
+        if all(file_dict.get("file_type") for file_dict in files):
             # if file_dict is retrieved from the API as `request_json`
             # first check if a directory can be created from the
             # "loose" files (alternative to parsing a .tar.gz file as
--- a/src/sparsezoo/model/utils.py
+++ b/src/sparsezoo/model/utils.py
@@ -18,6 +18,7 @@ import logging
 import os
 import re
 import shutil
+import tarfile
 import warnings
 from collections import defaultdict
 from pathlib import Path
@@ -576,11 +577,21 @@ def _copy_file_contents(
                 copy_path = os.path.join(output_dir, os.path.basename(_file.path))
                 _copy_and_overwrite(_file.path, copy_path, shutil.copyfile)
         elif isinstance(file, OnnxGz):
-            # copy all contents of unzipped onnx.tar.gz file to top level of output
-            onnx_gz_path = (
-                os.path.dirname(file.path) if os.path.isfile(file.path) else file.path
-            )
-            shutil.copytree(onnx_gz_path, output_dir, dirs_exist_ok=True)
+            # download and unzip model.onnx.tar.gz
+            model_gz_path = Path(file.path).with_name("model.onnx.tar.gz")
+
+            # copy all contents of unzipped model.onnx.tar.gz file to
+            #  top level of output
+            for name in tarfile.open(model_gz_path).getnames():
+                unzipped_file_path = model_gz_path.with_name(name)
+                copy_location = Path(output_dir) / name
+                _copy_and_overwrite(
+                    str(unzipped_file_path), str(copy_location), shutil.copyfile
+                )
+            # copy the model.onnx.tar.gz file to the top level of output
+            tar_copy_path = Path(output_dir) / model_gz_path.name
+            _copy_and_overwrite(str(model_gz_path), str(tar_copy_path), shutil.copyfile)
+
         elif isinstance(file, Directory):
             copy_path = os.path.join(output_dir, os.path.basename(file.path))
             _copy_and_overwrite(file.path, copy_path, shutil.copytree)
--- a/src/sparsezoo/objects/directories.py
+++ b/src/sparsezoo/objects/directories.py
@@ -369,4 +369,5 @@ class OnnxGz(Directory):
             if not unzipped_file_path.exists():
                 _LOGGER.info(f"{unzipped_file_path} does not exist, was it extracted?")
                 self.is_archive = True
+                return
         self.is_archive = False
--- a/src/sparsezoo/objects/directories.py
+++ b/src/sparsezoo/objects/directories.py
@@ -18,6 +18,7 @@ Class objects for standardization and validation of a model folder structure
 
 
 import logging
+import tarfile
 from collections import OrderedDict
 from pathlib import Path
 from typing import Dict, List, Optional, Union
@@ -285,15 +286,13 @@ class OnnxGz(Directory):
     with or without external data.
 
     Class Invariants:
-        - `self.name` attribute of this class will point to the name of the tarball
+        - `self.name` attribute of this class will point to the name of the onnx file
         - `self._path` and `self.path` will point to the path of the extracted
             onnx model
     """
 
     def __init__(self, *args, **kwargs):
-        name = kwargs.get("name")
         super().__init__(*args, **kwargs)
-        self.name = name
 
     def __repr__(self):
         return f"OnnxGz(name={self.name}, path={self._path})"
@@ -306,22 +305,31 @@ class OnnxGz(Directory):
                 name `model.onnx`
             - the tarball will be extracted to the same directory as the tarball
 
-        :post-condition: self._path will point to the path of the extrated
+        :post-condition: self._path will point to the path of the extracted
             onnx model, and it exists
         :return: path to the onnx model
         """
+
         expected_path: Path = (
             Path(self._path)
             if self._path is not None
-            else Path(self.parent_directory) / self.name
+            else Path(self.parent_directory) / "model.onnx.tar.gz"
         )
-        # point _path to model.onnx.tar.gz if first time else
-        #  model.onnx
+
+        # _path can also point to parent directory of the tarball when
+        #  this object is initialized from files
+        if expected_path.is_dir():
+            # point expected to model.onnx.tar.gz
+            expected_path = expected_path / "model.onnx.tar.gz"
+
+        # point path to model.onnx.tar.gz before download/unzip
         self._path = str(expected_path)
 
         if not expected_path.exists():
+            # download the tarball if it does not exist
             self.download()
 
+        self._check_if_extracted()
         if self.is_archive:
             # if the tarball is not extracted, extract it
             self.unzip()
@@ -336,3 +344,29 @@ class OnnxGz(Directory):
         # point _path to model.onnx
         self._path = str(onnx_model_path)
         return self._path
+
+    def _check_if_extracted(self):
+        """
+        set `is_archive` to False if the tarball is extracted, else set it to
+        True. Condition for being extracted is that all members of the tarball
+        are extracted and exist as files in the same directory as the tarball.
+        """
+        # expected file will point to the tarball
+        expected_path: Path = (
+            Path(self._path)
+            if self._path is not None
+            else Path(self.parent_directory) / "model.onnx.tar.gz"
+        )
+
+        if expected_path.is_dir():
+            # point expected to model.onnx.tar.gz
+            expected_path = expected_path / "model.onnx.tar.gz"
+
+        model_gz_path = expected_path.with_name(name="model.onnx.tar.gz")
+        # assert all members of  model.onnx.tar.gz have been extracted
+        for zipped_filename in tarfile.open(model_gz_path).getnames():
+            unzipped_file_path = expected_path.with_name(zipped_filename)
+            if not unzipped_file_path.exists():
+                _LOGGER.info(f"{unzipped_file_path} does not exist, was it extracted?")
+                self.is_archive = True
+        self.is_archive = False
--- a/src/sparsezoo/objects/directories.py
+++ b/src/sparsezoo/objects/directories.py
@@ -301,8 +301,9 @@ class OnnxGz(Directory):
     def path(self):
         """
         Assumptions:
-            - the tarball contains only one onnx model, with the
-                name `model.onnx`
+            - the tarball must contain atleast one `model.onnx` file
+            - the tarball may or may not contain additional external data or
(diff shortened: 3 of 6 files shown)


src/sparsezoo/model/model.py
```diff
@@ -157,7 +157,7 @@ def __init__(self, source: str, download_path: Optional[str] = None):
             stub_params=self.stub_params,
         )
 
-        self._onnx_gz: Directory = self._directory_from_files(
+        self._onnx_gz: OnnxGz = self._directory_from_files(
```
rahul-tuli, line 160: Changes in this file are merely cosmetic and could be a part of a separate PR

src/sparsezoo/model/model.py
```diff
@@ -587,12 +587,12 @@ def _directory_from_files(
         regex: Optional[bool] = False,
         allow_multiple_outputs: Optional[bool] = False,
         **kwargs: object,
-    ) -> Union[Directory, None]:
+    ) -> Union[List[Union[Directory, Any, None]], List[Directory], None]:
```
rahul-tuli, line 590: Fix typing

src/sparsezoo/model/utils.py
```diff
@@ -576,11 +577,21 @@ def _copy_file_contents(
                 copy_path = os.path.join(output_dir, os.path.basename(_file.path))
                 _copy_and_overwrite(_file.path, copy_path, shutil.copyfile)
         elif isinstance(file, OnnxGz):
-            # copy all contents of unzipped onnx.tar.gz file to top level of output
-            onnx_gz_path = (
-                os.path.dirname(file.path) if os.path.isfile(file.path) else file.path
-            )
-            shutil.copytree(onnx_gz_path, output_dir, dirs_exist_ok=True)
+            # download and unzip model.onnx.tar.gz
```
rahul-tuli, line 580: copy contents of OnnxGz file based on tarfile members rather than copying the whole parent directory

src/sparsezoo/objects/directories.py
```diff
@@ -283,14 +284,90 @@ class OnnxGz(Directory):
...
+    Class Invariants:
+        - `self.name` attribute of this class will point to the name of the onnx file
+        - `self._path` and `self.path` will point to the path of the extracted
+            onnx model
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def __repr__(self):
```
rahul-tuli, line 297: Add __repr__ for easier debugging

src/sparsezoo/objects/directories.py
```diff
@@ -283,14 +284,90 @@ class OnnxGz(Directory):
...
+            raise FileNotFoundError(
+                f"Expected to find model.onnx at {onnx_model_path}, "
+                "but it does not exist."
+            )
+
+        # point _path to model.onnx
+        self._path = str(onnx_model_path)
         return self._path
+
+    def _check_if_extracted(self):
```
rahul-tuli, line 348: checks if model.onnx.tar.gz was extracted based on all members inside the tarball, accordingly sets `is_archive`

src/sparsezoo/objects/directory.py
```diff
@@ -284,9 +284,15 @@ def unzip(self, extract_directory: Optional[str] = None, force: bool = False):
...
-
-        if not os.path.exists(path) or force:  # do not re-unzip if not forced
+        is_model_gz = name == "model.onnx"
+        path = (
+            extract_directory if is_model_gz else os.path.join(extract_directory, name)
+        )
+        # if is_model_gz then path would point to the tarfile(s)
+        # parent directory and must be unzipped irrespective of existence
+        if (
+            is_model_gz or not os.path.exists(path) or force
```
rahul-tuli, line 294: extract model.onnx.tar.gz

src/sparsezoo/objects/directory.py
```diff
@@ -340,7 +346,7 @@ def is_directory(file: File) -> bool:
 def _possibly_convert_files_to_directories(files: List[File]) -> List[File]:
     return [
         Directory.from_file(file)
-        if (is_directory(file) and not isinstance(file, Directory))
+        if not isinstance(file, Directory) and is_directory(file)
```
rahul-tuli, line 349: Move isinstance check first to avoid calling path property on Directory objects

src/sparsezoo/model/model.py
dsikka: `Any` doesn't cover the `None` case?

src/sparsezoo/model/utils.py
```diff
@@ -576,11 +577,21 @@ def _copy_file_contents(
...
+            #  top level of output
+            for name in tarfile.open(model_gz_path).getnames():
+                unzipped_file_path = model_gz_path.with_name(name)
+                copy_location = Path(output_dir) / name
+                _copy_and_overwrite(
+                    str(unzipped_file_path), str(copy_location), shutil.copyfile
+                )
+            # copy the model.onnx.tar.gz file to the top level of output
+            tar_copy_path = Path(output_dir) / model_gz_path.name
+            _copy_and_overwrite(str(model_gz_path), str(tar_copy_path), shutil.copyfile)
```
dsikka, line 593: why do you need to copy the tar file after already copying its untarred content in the for loop above?

src/sparsezoo/objects/directories.py
```diff
@@ -283,14 +284,90 @@ class OnnxGz(Directory):
...
+    def __repr__(self):
+        return f"OnnxGz(name={self.name}, path={self._path})"
+
     @property
     def path(self):
-        _ = super().path  # call self.path to download initial file if not already
+        """
+        Assumptions:
+            - the tarball contains only one onnx model, with the
+                name `model.onnx`
```
Satrat, line 305: This is not the case for the codegen models on sparsezoo, for at least one of them the tarball contains "model_nocache.onnx" Could we just search the directory for a .onnx file instead? We would still be assuming there is only one onnx model file in the folder

src/sparsezoo/model/model.py
rahul-tuli: Hmm, you are correct it should

src/sparsezoo/objects/directories.py
```diff
@@ -283,14 +284,90 @@ class OnnxGz(Directory):
...
+
+        self._check_if_extracted()
         if self.is_archive:
+            # if the tarball is not extracted, extract it
             self.unzip()
-        if os.path.isdir(self._path) and "model.onnx" in os.listdir(self._path):
-            # if unzipped into a directory, refer directly to model.onnx
-            self._path = os.path.join(self._path, "model.onnx")
+
+        onnx_model_path = expected_path.with_name(name="model.onnx")
```
Satrat, line 337: Do we need to enforce the model name as `model.onnx`? I think just searching for the file extension would be more flexible

src/sparsezoo/objects/directory.py
```diff
@@ -284,9 +284,15 @@ def unzip(self, extract_directory: Optional[str] = None, force: bool = False):
             )
 
         name = ".".join(self.name.split(".")[:-2])
-        path = os.path.join(extract_directory, name)
-
-        if not os.path.exists(path) or force:  # do not re-unzip if not forced
+        is_model_gz = name == "model.onnx"
```
Satrat, line 287: See other comments, not sure about hard coding this

src/sparsezoo/objects/directories.py
```diff
@@ -283,14 +284,90 @@ class OnnxGz(Directory):
...
+            expected_path = expected_path / "model.onnx.tar.gz"
+
+        model_gz_path = expected_path.with_name(name="model.onnx.tar.gz")
+        # assert all members of  model.onnx.tar.gz have been extracted
+        for zipped_filename in tarfile.open(model_gz_path).getnames():
+            unzipped_file_path = expected_path.with_name(zipped_filename)
+            if not unzipped_file_path.exists():
+                _LOGGER.info(f"{unzipped_file_path} does not exist, was it extracted?")
+                self.is_archive = True
+                return
```
Satrat, line 372: This is duplicated in another function, could we break it out into a helper?

Satrat (changes requested):
I'm pretty sure the file we want to unzip is model.tar.gz and not model.onnx.tar.gz if the purpose of this PR is to extract the model.data file.

[The 2B codegen stub](https://sparsezoo.neuralmagic.com/models/codegen_mono-2b-bigpython_bigquery_thepile-base?hardware=deepsparse-c6i.12xlarge&tab=4) shows model.tar.gz as 6.4GB and model.onnx.tar.gz as only 594KB.

In general, I think it would be a good idea to upload a small model to sparsezoo that has its data stored externally so we can catch the issue in the test suite without having to run giant models

src/sparsezoo/model/utils.py
rahul-tuli: There's no need per se; we can avoid copying over the tarball; from how I see it, it doesn't hurt or give us anything extra. I kept it in case someone wants to manually check it's contents, but I don't have a strong argument for either case.

src/sparsezoo/objects/directories.py
rahul-tuli: This is something we want to enforce with `model.onnx.tar.gz` (that it always contains a `model.onnx` along with external data files if any) I could see a future where we support other model names, but they would be handled separately and not via `model.onnx.tar.gz`, the codegen 2B model you mentioned has to be updated on sparsezoo

src/sparsezoo/objects/directories.py
rahul-tuli: I agree it would be more flexible, but let's leave this as is for now as its the only check enforcing constraints on the contents of `model.onnx.tar.gz` we can update this if the constraints change?

src/sparsezoo/objects/directory.py
rahul-tuli: This would work for now, as we are enforcing *decided* constraints on the contents of `model.onnx.tar.gz` but, am open to changing this logic if the constraints are ever relaxed

src/sparsezoo/objects/directories.py
rahul-tuli: That method uses this same iteration logic to copy/overwrite extracted files; only the iteration part is common and imo breaking the iteration part out might not be as beneficial. Thoughts? Happy to break this out if this is a hard blocker

rahul-tuli:
> I'm pretty sure the file we want to unzip is model.tar.gz and not model.onnx.tar.gz if the purpose of this PR is to extract the model.data file.

making a comment here that the linked model needs to be updated on sparsezoo; and the onnx model `model.onnx` along with external_data if any should be inside `model.onnx.tar.gz`

[The 2B codegen stub](https://sparsezoo.neuralmagic.com/models/codegen_mono-2b-bigpython_bigquery_thepile-base?hardware=deepsparse-c6i.12xlarge&tab=4) shows model.tar.gz as 6.4GB and model.onnx.tar.gz as only 594KB.

In general, I think it would be a good idea to upload a small model to sparsezoo that has its data stored externally so we can catch the issue in the test suite without having to run giant models

src/sparsezoo/objects/directories.py
Satrat: Makes sense, I'm all good to leave it

src/sparsezoo/model/utils.py
Satrat: My vote would be to avoid copying it, for large models especially we don't want to keep around unnecessary files

src/sparsezoo/model/model.py
rahul-tuli: But I believe it's nicer to make it explicit, will remove if it's a hard blocker for you

Satrat (approved):

src/sparsezoo/model/utils.py
rahul-tuli: Lol this function is only used in tests so I guess either way is fine

src/sparsezoo/objects/directories.py
dsikka: the other case where this may change quickly is for any pipeline requiring more than one model (encoder/decoder pipelines). We haven't come up with a definite plan for these on sparsezoo but I think *.onnx would be more generic?

src/sparsezoo/objects/directories.py
dsikka: you could do something where you check if multiple *.onnx files are present. If not, ensure the single one present is `model.onnx`

dsikka (commented):
Two comments. otherwise LGTM

src/sparsezoo/objects/directories.py
rahul-tuli: Let's leave it until we have a definite plan for these on sparsezoo; don't want to make any decisions here that are bound to change

src/sparsezoo/objects/directories.py
rahul-tuli: We would still need a definite plan on which of those onnx models should `Model(stub).onnx_model` point to; so I think this update would make more sense there?

src/sparsezoo/objects/directory.py
bfineran, line 294: why do we always unzip if `is_model_gz`? what would happen if the tarball still exists but has already been unzipped?

src/sparsezoo/objects/directories.py
```diff
@@ -283,14 +284,90 @@ class OnnxGz(Directory):
...
+        if expected_path.is_dir():
+            # point expected to model.onnx.tar.gz
+            expected_path = expected_path / "model.onnx.tar.gz"
+
+        model_gz_path = expected_path.with_name(name="model.onnx.tar.gz")
+        # assert all members of  model.onnx.tar.gz have been extracted
+        for zipped_filename in tarfile.open(model_gz_path).getnames():
+            unzipped_file_path = expected_path.with_name(zipped_filename)
+            if not unzipped_file_path.exists():
+                _LOGGER.info(f"{unzipped_file_path} does not exist, was it extracted?")
```
bfineran, line 370: should remove log - user can't do anything with this. potentially move to debug since this would be a weird state

src/sparsezoo/objects/directory.py
rahul-tuli: synced offline!

src/sparsezoo/objects/directories.py
rahul-tuli: Done!

bfineran (approved):

PR #358, opened 2023-08-18
Report GHA build results to Testmo

dhuangnm:
Improved GHA workflows to build nightly and release.

Added report build status to Testmo: https://neuralmagic.testmo.net/automation/sources/view/2


.github/workflows/build-nightly.yml
```diff
@@ -7,39 +7,16 @@ on:
...
-             fi
+
+    BUILD-SPARSEZOO_NIGHTLY:
+
+      uses: ./.github/workflows/util.yml
+      with:
+        runs_on: ubuntu-22.04
+        run_id: ${{ github.run_id }}
+        build_type: nightly
+        testmo_project_id: 9
```
dbarbuzzi, line 21: Shouldn’t this be `2` for sparsezoo? It seems we don’t have a Testmo project with an ID of 9, at least that I can see on my account.

.github/workflows/build-release.yml
```diff
@@ -5,32 +5,15 @@ on:
...
-             aws s3 cp log_${GITHUB_REF_NAME/\//-}_release_${todaytime}_${GITHUB_SHA:0:7} s3://nm-github-actions/${{ github.event.repository.name }}/
+
+    BUILD-SPARSEZOO_RELEASE:
+
+      uses: ./.github/workflows/util.yml
+      with:
+        runs_on: ubuntu-22.04
+        run_id: ${{ github.run_id }}
+        build_type: release
+        testmo_project_id: 9
```
dbarbuzzi, line 18: Same comment about confirming the right ID.

.github/workflows/util.yml
```diff
@@ -0,0 +1,124 @@
...
+          echo "  --name ${name} \\"
+          echo "  --source ${{ github.event.repository.name }} \\"
+          echo "  --results result.xml"
+          testmo automation:run:submit \
+                         --instance "${TESTMO_URL}" \
+                         --project-id ${{ inputs.testmo_project_id }} \
+                         --name "${name}" \
+                         --source "${{ github.event.repository.name }}" \
+                         --results result.xml \
+                         -- cat result.xml
```
dbarbuzzi, line 124: This will probably always report as a success (unless `result.xml` doesn’t exist) because it is using the return code from this command (`cat result.xml`) to determine success/failure.

dhuangnm: Yep good catch, updated to be able to report both success or failure.

.github/workflows/build-nightly.yml
dhuangnm: I created a new project Build which from testmo it looks its project id is 9.

bfineran (dismissed):

PR #365, opened 2023-09-12
RegistryMixin - tooling for easier registry/plugin patterns across NM repos

bfineran:
proposes a structure that can be used as a universal registry for NM parent classes that follow a registry pattern or future plugins

Classes or factories that require a registry can add `RegistryMixin` as a parent class to get access to the functionality. The included classmethods should work for each class that contains the mixin separately.

Supports either loading values either from the registered names or a `file:name` pattern for custom file impelementations.

brief example from doc
```python
class Dataset(RegistryMixin):
    pass


# register with default name
@Dataset.register()
class ImageNetDataset(Dataset):
    pass

# load as "ImageNetDataset"
imagenet = Dataset.load("ImageNetDataset")

# register with custom name
@Dataset.register(name="cifar-dataset")
class Cifar(Dataset):
    pass

# load as "cifar-dataset"
cifar = Dataset.load_from_registry("cifar-dataset")

# load from custom file that implements a dataset
mnist = Dataset.load_from_registry("/path/to/mnnist_dataset.py:MnistDataset")
```

**test_plan:**
simple unit tests included

### DIFF
--- a/src/sparsezoo/utils/registry.py
+++ b/src/sparsezoo/utils/registry.py
@@ -1,0 +1,168 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Universal registry to support registration and loading of child classes and plugins
+of neuralmagic utilities
+"""
+
+import importlib
+from collections import defaultdict
+from typing import Any, Dict, Optional, Type
+
+
+_REGISTRY: Dict[Type, Dict[str, Any]] = defaultdict(dict)
+
+
+class RegistryMixin:
+    """
+    Universal registry to support registration and loading of child classes and plugins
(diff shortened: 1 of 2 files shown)


Satrat (approved):

src/sparsezoo/utils/registry.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+    return value
+
+
+def _validate_subclass(parent_class: Type, child_class: Type):
+    if not issubclass(child_class, parent_class):
+        raise ValueError(
+            f"class {child_class} is not a subclass of the class it is "
+            f"registered for: {parent_class}."
+        )
```
rahul-tuli, line 242:
```suggestion
        raise ValueError(
            f"class {child_class.__qualname__} is not a subclass of the class it is "
            f"registered for: {parent_class.__qualname__}."
        )
```

Else we get an uglier print message like:
`ValueError: class <class '__main__.B'> is not a subclass of the class it is registered for: <class '__main__.A'>.`

rahul-tuli (dismissed):

src/sparsezoo/utils/registry.py
```diff
@@ -0,0 +1,242 @@
...
+    def register(cls, name: Optional[str] = None):
+        """
+        Decorator for registering a value (ie class or function) wrapped by this
+        decorator to the base class (class that .register is called from)
+
+        :param name: name to register the wrapped value as, defaults to value.__name__
+        :return: register decorator
+        """
+
+        def decorator(value: Any):
```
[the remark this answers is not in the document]
bfineran: internal function to denote it's a returned decorator, not sure elaboration is needed here

src/sparsezoo/utils/registry.py
bfineran: using the class for the printout so users will see the exact class rather than name for error message, useful for overlapping names

PR #366, opened 2023-09-14
Restore "default" Property for Recipes

Satrat:
With the SparseZoo V2 refactor, model recipes are stored in the root directory rather than a subfolder, so the Model class stored them as a List[Files] rather than a Directory. This removed the functionality to be able to reference a default recipe with recipes.default.

This PR introduces a Recipes class that restores the recipes.default functionality. PR https://github.com/neuralmagic/sparseml/pull/1719 will have to be updated after this is merged to reflect the change

### DIFF
--- a/src/sparsezoo/objects/recipes.py
+++ b/src/sparsezoo/objects/recipes.py
@@ -47,7 +47,7 @@ class Recipes:
         self._default_recipe_name = self._RECIPE_DEFAULT_NAME
         custom_default = stub_params.get("recipe_type") or stub_params.get("recipe")
         if custom_default is not None:
-            self._default_recipe_name = "recipe_" + custom_default + ".md"
+            self._default_recipe_name = "recipe_" + custom_default
 
     @property
     def recipes(self) -> List:
@@ -62,7 +62,7 @@ class Recipes:
         :return: The default recipe in the recipe list
         """
         for recipe in self._recipes:
-            if recipe.name == self._default_recipe_name:
+            if recipe.name.startswith(self._default_recipe_name):
                 return recipe
 
         # fallback to first recipe in list
--- a/src/sparsezoo/objects/recipes.py
+++ b/src/sparsezoo/objects/recipes.py
@@ -28,7 +28,7 @@ class Recipes:
     Object to store a list of recipes for a downloaded model and pull the default
 
     :param recipes: list of recipes to store
-    :param stub_params: dictionary storing custom default recipes names
+    :param stub_params: dictionary that may contain custom default recipes names
     """
 
     _RECIPE_DEFAULT_NAME = "recipe.md"
--- a/src/sparsezoo/objects/recipes.py
+++ b/src/sparsezoo/objects/recipes.py
@@ -32,7 +32,9 @@ class Recipes:
 
     _RECIPE_DEFAULT_NAME = "recipe.md"
 
-    def __init__(self, recipes: Union[File, List[File]]):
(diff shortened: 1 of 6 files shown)


horheynm (dismissed):

tests/sparsezoo/objects/test_recipes.py
```diff
@@ -0,0 +1,55 @@
...
+
+    found_by_name = model.recipes.get_recipe_by_name("does_not_exist.md")
+    assert found_by_name is None
+
+    shutil.rmtree(temp_dir.name)
+
+
+def test_custom_default():
+    custom_default_name = "transfer_text_classification"
+    stub_with_multiple_recieps = (
```
horheynm, line 44: typo

tests/sparsezoo/objects/test_recipes.py
```diff
@@ -0,0 +1,55 @@
...
+            custom_default_name
+        )
+    )
+    temp_dir = tempfile.TemporaryDirectory(dir="/tmp")
+    model = Model(stub_with_multiple_recieps, temp_dir.name)
+
+    expected_default_name = "recipe_" + custom_default_name + ".md"
+
+    default_recipe = model.recipes.default
+    assert default_recipe.name == expected_default_name
```
horheynm, line 55: delete tmp dir?
Might be worth making a fixture that yields tmp dir and deletes after

dbarbuzzi: Pytest actually has that built-in: https://docs.pytest.org/en/7.4.x/how-to/tmp_path.html

src/sparsezoo/objects/recipes.py
```diff
@@ -0,0 +1,85 @@
...
+    Object to store a list of recipes for a downloaded model and pull the default
+
+    :param recipes: list of recipes to store
+    :param stub_params: dictionary that may contain custom default recipes names
+    """
+
+    _RECIPE_DEFAULT_NAME = "recipe.md"
+
+    def __init__(
+        self, recipes: Union[None, File, List[File]], stub_params: Dict[str, str]
```
horheynm, line 37: Union[None,...] is same as Optional[Union[F, L[F]] = None
and i would make stub_params: Dict[str, str] = {}

src/sparsezoo/objects/recipes.py
```diff
@@ -0,0 +1,85 @@
...
+    ):
+        if recipes is None:
+            recipes = []
+        if isinstance(recipes, File):
+            recipes = [recipes]
+        self._recipes = recipes
+
+        self._default_recipe_name = self._RECIPE_DEFAULT_NAME
+        custom_default = stub_params.get("recipe_type") or stub_params.get("recipe")
```
bfineran, line 46: I think we also would want to match to `recipe.name.startswith(custom_default)`

Satrat: Recipes seem to be required to start with "recipe_" so I don't think we need to check that here. `?recipe=testname` matches to `recipe_testname` when filtering recipes, see https://github.com/neuralmagic/sparsezoo/blob/main/src/sparsezoo/model/utils.py#L230.

@horheynm thoughts?

Satrat: Ended up dealing with this by removing the ".md" extension for custom defaults, now we can check for matches using startswith

bfineran (approved):

horheynm (approved):

PR #368, opened 2023-09-22
Add  "available" attribute to recipes

Satrat:
Fixes the error in this Asana ticket: https://app.asana.com/0/1204322589046915/1205549265505143/f

Adds an `available` attribute to `Recipes` that provides a list of recipe names. This is needed by the `sparsify.run` sparse transfer

### Testing
Confirmed the command failing in the ticket runs:
```
sparsify.run sparse-transfer --use-case image-classification --data imagenette --optim-level 0.5
```
Also updated unit tests to test the new property

### DIFF
--- a/src/sparsezoo/objects/recipes.py
+++ b/src/sparsezoo/objects/recipes.py
@@ -76,6 +76,9 @@ class Recipes:
                 return recipe
 
         # fallback to first recipe in list
+        if len(self._recipes) == 0:
+            raise ValueError("No recipes found, could not retrieve a default.")
+
         _LOGGER.warning(
             f"No default recipe {self._default_recipe_name} found, falling back to "
             f"first listed recipe {self._recipes[0].name}"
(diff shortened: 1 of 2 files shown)


src/sparsezoo/objects/recipes.py
```diff
@@ -67,8 +77,8 @@ def default(self) -> File:
 
         # fallback to first recipe in list
         _LOGGER.warning(
-            "No default recipe {self._default_recipe_name} found, falling back to"
-            "first listed recipe"
+            f"No default recipe {self._default_recipe_name} found, falling back to "
+            f"first listed recipe {self._recipes[0].name}"
```
rahul-tuli, line 81: Could we add an assert or Raise an exception if self._recipes is empty?

rahul-tuli (approved):
Nice clean diff!

src/sparsezoo/objects/recipes.py
Satrat: Yup, done!

rahul-tuli (approved):

bfineran (approved):

PR #370, opened 2023-09-26
Use file download URL as provided by the API

tdg5:
Sparsezoo shouldn't be generating file download URLs, it should use the file downloads provided by the API. The download URLs from the API are cheap and allow sparsezoo to be agnostic of where the files live on the interwebs and decoupled from whatever the current file download implementation is.

### DIFF
--- a/src/sparsezoo/model/utils.py
+++ b/src/sparsezoo/model/utils.py
@@ -599,9 +599,8 @@ def _copy_and_overwrite(from_path, to_path, func):
 
 def include_file_download_url(files: List[Dict]):
     for file in files:
-        file["url"] = get_file_download_url(
-            model_id=file["model_id"], file_name=file["display_name"]
-        )
+        file["url"] = get_file_download_url(file["download_url"])
+        del file["download_url"]
 
 
 def get_model_metadata_from_stub(stub: str) -> Dict[str, str]:
@@ -637,15 +636,12 @@ def is_stub(candidate: str) -> bool:
 
 
 def get_file_download_url(
-    model_id: str,
-    file_name: str,
-    base_url: str = BASE_API_URL,
+    download_url: str,
 ):
     """Url to download a file"""
-    download_url = f"{base_url}/v2/models/{model_id}/files/{file_name}"
-
     # important, do not remove
     if convert_to_bool(os.getenv("SPARSEZOO_TEST_MODE")):
-        download_url += "?increment_download=False"
+        delimiter = "&" if "?" in download_url else "?"
+        download_url += f"{delimiter}increment_download=False"
 
     return download_url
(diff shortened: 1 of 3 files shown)


horheynm (approved):

src/sparsezoo/model/utils.py
```diff
@@ -637,15 +636,12 @@ def is_stub(candidate: str) -> bool:
...
-    base_url: str = BASE_API_URL,
+    download_url: str,
 ):
     """Url to download a file"""
-    download_url = f"{base_url}/v2/models/{model_id}/files/{file_name}"
-
     # important, do not remove
     if convert_to_bool(os.getenv("SPARSEZOO_TEST_MODE")):
-        download_url += "?increment_download=False"
+        delimiter = "&" if "?" in download_url else "?"
```
bfineran, line 644: nice

src/sparsezoo/model/utils.py
```diff
@@ -599,9 +599,8 @@ def _copy_and_overwrite(from_path, to_path, func):
 
 def include_file_download_url(files: List[Dict]):
     for file in files:
-        file["url"] = get_file_download_url(
-            model_id=file["model_id"], file_name=file["display_name"]
-        )
+        file["url"] = get_file_download_url(file["download_url"])
```
bfineran, line 602: where is this used now?

src/sparsezoo/model/utils.py
```diff
@@ -599,9 +599,8 @@ def _copy_and_overwrite(from_path, to_path, func):
...
+        del file["download_url"]
```
bfineran, line 603: why delete?

src/sparsezoo/model/utils.py
bfineran: actually, never mind, see it was replacing the previous key

tdg5: The `get_file_download_url` function is only called from this file, I can get rid of that method and move the logic here if you like. That would look something like:
```python
def include_file_download_url(files: List[Dict]):
    for file in files:
        download_url = file["download_url"]
        del file["download_url"]

        # important, do not remove
        if convert_to_bool(os.getenv("SPARSEZOO_TEST_MODE")):
            delimiter = "&" if "?" in download_url else "?"
            download_url += f"{delimiter}increment_download=False"

        file["url"] = download_url
```

src/sparsezoo/model/utils.py
tdg5: Not required, just trying to maintain the existing file dict interface of sorts. I'd think having both a `url` and `download_url` property would confuse anyone looking at the dict and trying to make sense of things

src/sparsezoo/model/utils.py
bfineran: no need, looks good thanks

bfineran (approved):

PR #378, opened 2023-10-24
[Wip] Fix aliased directory

dbogunowicz:


src/sparsezoo/objects/directories.py
```diff
@@ -302,48 +302,31 @@ def __init__(self, *args, download_alias: Optional[str] = None, **kwargs):
...
-        )
-        if not actual_file_path.exists():
-            raise FileNotFoundError(
-                f"The directory {actual_file_path} should have been "
-                "downloaded but does not exist"
-            )
-        return str(actual_file_path)
+        with self._download_with_alias():
+            super().download(*args, **kwargs)
+            self.unzip()
```
bfineran, line 329: why are we moving this functionality here? this shouldn't be a requirement/side effect of an "aliased" directory

PR #379, opened 2023-10-24
Enable the easy download of the deployment.tar.gz

dbogunowicz:
Reverting to the state of #375 in order to enable, in the simplest way possible, the fetching of only `deployment.tar.gz`.
The crux of this PR is adding `self.deployment_tar` attribute to the `Model` and allowing it to be correctly initialized not only from the stub but also from the local directory.
Added unit tests that test this feature.

Manual tests:

```python
path_where_models_are = "/home/ubuntu/.cache/sparsezoo/neuralmagic/codegen_mono-350m-bigpython_bigquery_thepile-pruned50_quantized"
if os.path.exists(path_where_models_are):
    shutil.rmtree(path_where_models_are)
    print(f"The path {path_where_models_are} has been cleaned up")
else:
    print(f"The path {path_where_models_are} does not exist")


print("1: Creating the deployment from stub")
model = Model("zoo:codegen_mono-350m-bigpython_bigquery_thepile-pruned50_quantized")
# local path to the `deployment.tar.gz` should be None (we have not downloaded anything yet)
print("Deployment directory tar path (before download): ", model.deployment_tar._path)
# grab the deployment_dir_path (this downloads the deployment.tar.gz and unpacks it)
deployment_dir_path = model.deployment_directory_path

# all should be pointing to `directory`
assert deployment_dir_path == model.deployment_tar.path == model.deployment.path
print("Deployment directory tar path: ", model.deployment_tar._path)
print("Deployment directory path: ", model.deployment.path)
print(f"deployment_dir_path: {deployment_dir_path}")

# all should display `directory` files
print("Contents of the deployment tar directory: ", os.listdir(model.deployment_tar.path))
print("Contents of the deployment directory: ", os.listdir(model.deployment.path))

# should only contain `deployment` and `deployment.tar.gz`
print("Contents of the model directory: ", os.listdir(model._path))

print("2: Creating the deployment from local path")
model = Model("zoo:codegen_mono-350m-bigpython_bigquery_thepile-pruned50_quantized")
print("Deployment directory tar path: ", model.deployment_tar.path)
deployment_dir_path = model.deployment_directory_path

assert deployment_dir_path == model.deployment_tar.path
print("Deployment directory tar path (after unzipping): ", model.deployment_tar._path)
print("Deployment directory path: ", model.deployment.path)
print(f"deployment_dir_path: {deployment_dir_path}")
print("Contents of the deployment tar directory: ", os.listdir(model.deployment_tar.path))
[...]

PR #386, opened 2023-11-15
[Cherry-Pick] Enable the easy download of the deployment.tar.gz

dbogunowicz:
Cherry picking https://github.com/neuralmagic/sparsezoo/pull/379 to `release/1.6`. 
The goal is to mitigate the problem with repeated downloads of the `directory` folder, caused by the simultaneous presence of the `deployment` directory and `deployment.tar.gz` file.


bfineran:
duplicated by #387

PR #394, opened 2023-11-29
Parameter Analysis

horheynm:
```python3
import onnx

from sparsezoo.utils import (
    ONNXGraph,
    extract_node_id,
    extract_node_shapes_and_dtypes,
    get_node_num_zeros_and_size,
)
from sparsezoo.analyze_v2 import ParameterAnalysis

path = "/Users/george/.cache/sparsezoo/neuralmagic/resnet_v1-50-imagenet-pruned95_uniform_quantized/deployment/model.onnx"

model = onnx.load(path)
model_graph = ONNXGraph(model)
node_shapes, node_dtypes = extract_node_shapes_and_dtypes(model_graph.model)
parameter_analysis = {}
for node in model_graph.nodes:
    node_id = extract_node_id(node)
    node_shape = node_shapes.get(node_id)
    node_dtype = node_dtypes.get(node_id)
    
    parameter_analysis[node_id] = ParameterAnalysis(model_graph, node)
    # print(parameter_analysis[node_id].to_dict())
    # print(parameter_analysis[node_id].to_yaml())
```

### DIFF
--- a/src/sparsezoo/analyze_v2/model_validators/distribution_analysis_model.py
+++ b/src/sparsezoo/analyze_v2/model_validators/distribution_analysis_model.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict, List, Optional, Union
+
+from pydantic import BaseModel, validator
+
+from sparsezoo.analyze_v2.model_validators.utils import type_validator
+
+
+class DistributionAnalysisModel(BaseModel):
+    counts: Optional[int]
+    mean: Optional[float]
+    median: Optional[float]
+    # modes: Optional[Union[List[float], numpy.ndarray]]
+    modes: Optional[List]
+    sum_val: Optional[float]
+    min_val: Optional[float]
+    max_val: Optional[float]
+    percentiles: Optional[Dict[float, float]]  # 0.1, 0.25, 0.5, 0.75, 0.9
+    std_dev: Optional[float]
+    skewness: Optional[float]
+    kurtosis: Optional[float]
+    entropy: Optional[float]
+    bin_width: Optional[float]  # (max_val - min_val) / num_bins
+    num_bins: Optional[int]  # 25
+    hist: Optional[List[float]]  # counts in each bin
+    bin_edges: Optional[List[float]]  # edge value for each bin
+
+    @validator("*", pre=True)
+    def validate_types(cls, value):
+        return type_validator(value)
--- a/src/sparsezoo/analyze_v2/parameter_analysis.py
+++ b/src/sparsezoo/analyze_v2/parameter_analysis.py
@@ -1,0 +1,208 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict, Union
+
+import numpy
+import yaml
+from onnx import NodeProto
+
+from sparsezoo.analyze_v2.model_validators import (
+    DistributionAnalysisModel,
+    QuantizationAnalysisModel,
+    SparsityAnalysisModel,
+)
+from sparsezoo.utils import (
+    ONNXGraph,
+    get_node_num_four_block_zeros_and_size,
+    get_node_param_counts,
+    get_node_weight,
+    get_node_weight_bits,
+    get_numpy_distribution_statistics,
+    get_numpy_entropy,
+    get_numpy_modes,
+    get_numpy_percentiles,
+    is_quantized_layer,
+)
+
+
+class ParameterAnalysis:
+    """
+    Given model_graph and node, compute
+
+    1. counts / sparsity level
+    2. bits / quantization level
+    3. distribution
+
+    with respect to the weights in the node, if any.
+    """
+
+    def __init__(
+        self,
+        model_graph: ONNXGraph,
+        node: NodeProto,
+    ):
+        self.model_graph = model_graph
+        self.node = node
+
+        self.counts = self.get_counts()
+        self.bits = self.get_bits()
+        self.distribution = self.get_distribution()
+
+    def get_counts(self) -> Dict[str, Union[int, float]]:
+        """Get the number of dense and sparse weights"""
+        data = get_parameter_counts(self.model_graph, self.node)
+        return {
+            grouping: SparsityAnalysisModel(**counts_dict).dict()
+            for grouping, counts_dict in data.items()
+        }
+
+    def get_bits(self) -> Dict[str, Union[int, float]]:
+        """Get the number of bits and quantized bits from weights"""
+        data = get_parameter_bits(self.model_graph, self.node)
+        return {
+            grouping: QuantizationAnalysisModel(**bits_dict).dict()
+            for grouping, bits_dict in data.items()
+        }
+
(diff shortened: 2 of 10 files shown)


src/sparsezoo/analyze_v2/model_validators/distribution_analysis_model.py
```diff
@@ -0,0 +1,42 @@
...
+    std_dev: Optional[float]
+    skewness: Optional[float]
+    kurtosis: Optional[float]
+    entropy: Optional[float]
+    bin_width: Optional[float]
+    num_bins: Optional[int]
+    hist: Optional[List[float]]
+    bin_edges: Optional[List[float]]
+
+    @validator("*", pre=True)
```
dbogunowicz, line 40: What is the role of the asterisk and pre=True in this validator method?

src/sparsezoo/analyze_v2/model_validators/distribution_analysis_model.py
```diff
@@ -0,0 +1,42 @@
...
+
+from typing import Dict, List, Optional, Union
+
+from pydantic import BaseModel, validator
+
+from sparsezoo.analyze_v2.model_validators.utils import type_validator
+
+
+class DistributionAnalysisModel(BaseModel):
+    counts: Optional[int]
```
dbogunowicz, line 23: Let's add some documentation in form of `= Field(description= "...")`

src/sparsezoo/utils/onnx/analysis.py
```diff
@@ -438,3 +445,82 @@ def _get_node_input(
...
+def get_numpy_modes(arr: numpy.ndarray):
+    unique_values, counts = numpy.unique(arr.flatten(), return_counts=True)
+    max_value_indices = numpy.where(counts == numpy.max(counts))
+    mode = unique_values[max_value_indices]
+    return mode.tolist()
+
+
+def get_numpy_percentiles(
+    arr: numpy.ndarray, percentiles: List[int] = [10, 25, 50, 75, 90]
+):
```
dbogunowicz, line 518: nit: return types are missing across functions

src/sparsezoo/utils/onnx/analysis.py
```diff
@@ -438,3 +445,82 @@ def _get_node_input(
...
+    skewness = (numpy.sum((flatten_arr - mean_val) ** 3) / n) / (std_dev ** 3)
+    kurtosis = (numpy.sum((flatten_arr - mean_val) ** 4) / n) / (std_dev ** 4) - 3
+
+    return skewness, kurtosis
+
+
+def get_numpy_entropy(arr: numpy.ndarray):
+    """Remove dimensions and compute the overall entropy"""
+    flatten_arr = arr.flatten()
+    flatten_arr = flatten_arr[flatten_arr > 0]
```
dbogunowicz, line 503: why are we removing negative values?

src/sparsezoo/analyze_v2/parameter_analysis.py
```diff
@@ -0,0 +1,208 @@
...
+            "counts": num_weights,
+            "counts_sparse": num_weights_sparse,
+            "percent": num_weights_sparse / num_weights if num_weights > 0 else 0,
+        },
+        "block4": {
+            "counts": num_weights_four_block,
+            "counts_sparse": num_sparse_weights_four_blocks,
+            "percent": num_sparse_weights_four_blocks / num_weights_four_block
+            if num_weights > 0
+            else 9,
```
dbogunowicz, line 122: why nine? such a random number :)

src/sparsezoo/analyze_v2/parameter_analysis.py
```diff
@@ -0,0 +1,208 @@
...
+        bits = get_node_weight_bits(model_graph, node)
+
+    bits_quant = bits * is_quantized_layer(model_graph, node)
+    return {
+        "tensor": {
+            "bits": bits,
+            "bits_quant": bits * is_quantized_layer(model_graph, node),
+            "percent": bits_quant / bits if bits > 0 else 0,
+        },
+        # TODO: Channels
```
dbogunowicz, line 150: Intended?

dbogunowicz (changes requested):
Great job overall. I especially love the clear structure an the "bite-sized" PR

src/sparsezoo/analyze_v2/model_validators/distribution_analysis_model.py
horheynm: "*" tells pydantic to run the validator for all values. 
The reason I do it for all is because yaml conversion will be messed up in `to_yaml()`, because some type might be numpy.generic.
 pre tells to run the validator before populating the values

src/sparsezoo/analyze_v2/model_validators/distribution_analysis_model.py
```diff
@@ -0,0 +1,42 @@
...
+    mean: Optional[float]
+    median: Optional[float]
```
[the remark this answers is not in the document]
horheynm: because some nodes may not have weights, so instead of outputting zero, which means there is a value, i made it None, so when we compute the aggregated values it doesn't consider None as a value

src/sparsezoo/analyze_v2/parameter_analysis.py
```diff
@@ -0,0 +1,208 @@
...
+        self.node = node
+
+        self.counts = self.get_counts()
+        self.bits = self.get_bits()
+        self.distribution = self.get_distribution()
+
+    def get_counts(self) -> Dict[str, Union[int, float]]:
+        """Get the number of dense and sparse weights"""
+        data = get_parameter_counts(self.model_graph, self.node)
+        return {
```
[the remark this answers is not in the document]
horheynm: sure, i dont mind. Are there attributes in basemodel you were think of using in the future from this output?

src/sparsezoo/utils/onnx/analysis.py
horheynm: This gives Nan or some non number value from the equation if i remember right

src/sparsezoo/analyze_v2/parameter_analysis.py
horheynm: oh nice catch

src/sparsezoo/analyze_v2/parameter_analysis.py
horheynm: Yes, because not needed for sparsezoo ui for now. We'll need it for sparsify.

src/sparsezoo/analyze_v2/model_validators/distribution_analysis_model.py
horheynm: Ok

dbogunowicz (approved):

src/sparsezoo/analyze_v2/model_validator/parameter_analysis_model.py
```diff
@@ -0,0 +1,35 @@
...
+)
+from sparsezoo.analyze_v2.model_validator.quantization_analysis_model import (
+    QuantizationAnalysisModel,
+)
+from sparsezoo.analyze_v2.model_validator.sparsity_analysis_model import (
+    SparsityAnalysisModel,
+)
+
+
+class ParameterAnalysisModel(BaseModel):
```
bfineran, line 30: we don't need `Model` in all the subclass names since the rest tells us what it really is and the base pydantic class is more there for convienence (ie we wouldn't name a dataclass as `_DataClass`)

src/sparsezoo/analyze_v2/parameter_analysis.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self.counts: Dict[str, int] = {}  # single grouping param counts
+        self.bits: Dict[str, int] = {}  # Tensor grouping bits
+
+        self.sparsity_analysis_model = self.get_sparsity()
+        self.quantization_analysis_model = self.get_quantization()
+        self.distribution_model = self.get_distribution()
+
+    def get_sparsity(self) -> List["SparsityAnalysisModel"]:
+        """Get the number of dense and sparse weights"""
```
bfineran, line 68: we need to standardize your docstrings to match repo standards

```
"""
[Generic info] (optional if covered by the rest)

:param PARAM1: ... (don't include self)
:param PARAM2: ...
:returns: ...
"""
```

src/sparsezoo/analyze_v2/parameter_analysis.py
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+
+def get_parameter_distribution(
+    model_graph: ONNXGraph,
+    node: NodeProto,
+    num_bins: int = 25,
+    *args,
+    **kwargs,
+) -> Dict[str, Union[int, float]]:
```
bfineran, line 170: why not just return the model?

src/sparsezoo/analyze_v2/parameter_analysis.py
```diff
@@ -0,0 +1,214 @@
...
+            quantization_analysis_model.append(
+                QuantizationAnalysisModel(grouping=grouping, **counts_dict)
+            )
+
+        return quantization_analysis_model
+
+    def get_distribution(self) -> "DistributionAnalysisModel":
+        """Get the distribution statistics with respect to the weights"""
+        distribution_dct = get_parameter_distribution(self.model_graph, self.node)
+        return DistributionAnalysisModel(**distribution_dct)
```
bfineran, line 99: why not just return `None` if there's no weight. I think this will make it easier to check, rather than a user getting a constructed object with all the attrs set to None

bfineran (approved):

PR #404, opened 2023-12-05
`RegistryMixin` improved alias management

dbogunowicz:
## Feature Description

Adds additional features the the `RegistryMixin`:
1. Separates concept of `name` and `alias`. Now, we can only have one name and several aliases
2. Registered names and aliases can be independently inspected using the appropriate methods
3. Registering and looking up names will be now robust wrt underscores and spaces (see the docstrings and tests)
4. Improves tests so their are more readable and extendable

### DIFF
--- a/src/sparsezoo/utils/registry.py
+++ b/src/sparsezoo/utils/registry.py
@@ -225,8 +225,9 @@ def get_from_registry(
         module_path, value_name = name.split(":")
         retrieved_value = _import_and_get_value_from_module(module_path, value_name)
     else:
-        # look up name in registry
+        # look up name in alias registry
         name = _ALIAS_REGISTRY.get(name)
+        # look up name in registry
         retrieved_value = _REGISTRY[parent_class].get(name)
         if retrieved_value is None:
             raise KeyError(
@@ -293,6 +294,12 @@ def register_alias(name: str, alias: Union[str, List[str], None] = None):
     alias = _add_hyphen_underscores_variants(alias)
 
     for alias_name in alias:
+        if alias_name in _ALIAS_REGISTRY:
+            raise KeyError(
+                f"Attempting to register alias {alias_name} as {name} "
+                f"however {alias_name} has already been registered as "
+                f"{_ALIAS_REGISTRY[alias_name]}"
+            )
         _ALIAS_REGISTRY[alias_name] = name
 
 
--- a/src/sparsezoo/utils/registry.py
+++ b/src/sparsezoo/utils/registry.py
@@ -67,7 +67,7 @@ class RegistryMixin:
         pass
 
     # register with multiple aliases
-    @Dataset.register(name=["cifar-10-dataset", "cifar-100-dataset"])
+    @Dataset.register(alias=["cifar-10-dataset", "cifar_100_dataset"])
     class Cifar(Dataset):
         pass
 
@@ -77,6 +77,13 @@ class RegistryMixin:
     # load from custom file that implements a dataset
     mnist = Dataset.load_from_registry("/path/to/mnnist_dataset.py:MnistDataset")
     ```
+
+    Note, that any name or alias that is being registered, will be also recognized
+    when all hyphens are replaced with underscores and vice versa.
+    For example, if a class is registered:
+     - as "cifar-10-dataset", it will be also recognized as "cifar_10_dataset"
+     - as "cifar_10_dataset", it will be also recognized as "cifar-10-dataset"
+        etc.
     """
 
     # set to True in child class to add check that registered/retrieved values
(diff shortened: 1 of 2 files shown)


src/sparsezoo/utils/registry.py
```diff
@@ -223,6 +251,51 @@ def registered_names(parent_class: Type) -> List[str]:
...
+    """
+    if alias is not None:
+        alias = alias if isinstance(alias, list) else [alias]
+    else:
+        alias = []
+    alias.append(name)
+    alias = _add_hyphen_underscores_variants(alias)
+
+    for alias_name in alias:
+        _ALIAS_REGISTRY[alias_name] = name
```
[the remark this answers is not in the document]
dbogunowicz: what if a person uses the name as the alias? cast in the outer function perhaps.

src/sparsezoo/utils/registry.py
```diff
@@ -27,12 +27,32 @@
...
+
+    example:
+    ```
+    standardize_lookup_name("foo_bar baz") == "foo-bar-baz"
+    ```
+
+    :param name: name to standardize
+    :return: standardized name
+    """
+    return name.replace("_", "-").replace(" ", "-")
```
bfineran, line 53: let's also call `.lower()` and strip whitespace, maybe `.`s as well...

src/sparsezoo/utils/registry.py
```diff
@@ -149,24 +181,37 @@ def registered_names(cls) -> List[str]:
...
+        defaults to None
     :param require_subclass: require that value is a subclass of the class this
         method is called from
     """
     if name is None:
         # default name
         name = value.__name__
 
+    name = standardize_lookup_name(name)
+    register_alias(name=name, alias=alias, parent_class=parent_class)
```
bfineran, line 213: we need to standardize the aliases as well right?

bfineran, line 213: should also as ease of use filter any aliases that standardize to the same str

src/sparsezoo/utils/registry.py
```diff
@@ -64,10 +84,16 @@ class ImageNetDataset(Dataset):
...
         pass
 
+    Note: the name will be standardized for lookup in the registry.
+    For example, if a class is registered as "cifar_dataset" or
+    "cifar dataset", it will be stored as "cifar-dataset". The user
+    will be able to load the class with any of the three name variants.
+
     # register with multiple aliases
-    @Dataset.register(name=["cifar-10-dataset", "cifar-100-dataset"])
+    @Dataset.register(alias=["cifar-10-dataset", "cifar_100_dataset"])
```
Satrat, line 93: I think I'm confused here, where does the actual registry name get set? Is "datasetname-dataset" a default if a name isn't provided?

src/sparsezoo/utils/registry.py
```diff
@@ -27,12 +27,32 @@
...
 
+_ALIAS_REGISTRY: Dict[Type, Dict[str, str]] = defaultdict(dict)
 _REGISTRY: Dict[Type, Dict[str, Any]] = defaultdict(dict)
 
 
+def standardize_lookup_name(name: str) -> str:
+    """
+    Standardize the given name for lookup in the registry.
+    This will replace all underscores and spaces with hyphens and
+    convert the name to lowercase.
```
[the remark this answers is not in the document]
dbogunowicz: good catch!

src/sparsezoo/utils/registry.py
dbogunowicz: Do we? I thought that I'd be nice (hopefully) not confusing aliases to be more "liberal", i.e. the user can set them directly to their face value, but I do not have a strong opinion there.

src/sparsezoo/utils/registry.py
dbogunowicz: in this case the name would be `Cifar`

src/sparsezoo/utils/registry.py
Satrat: I'd vote to be consistent and standardize everything, if we standardize the inputs as well then the user wouldn't notice a change right?

Satrat (dismissed):
LGTM, would just vote to standardize aliases in addition to names!

src/sparsezoo/utils/registry.py
```diff
@@ -194,19 +239,24 @@ def get_from_registry(
...
     """
+    name = standardize_lookup_name(name)
 
     if ":" in name:
         # user specifying specific module to load and value to import
         module_path, value_name = name.split(":")
         retrieved_value = _import_and_get_value_from_module(module_path, value_name)
     else:
+        # look up name in alias registry
+        name = _ALIAS_REGISTRY[parent_class].get(name)
```
Satrat, line 250: actually don't we need to standardize aliases? Because on line 242 we're standardizing the input, so we wouldn't match to an unstandardized alias

bfineran (approved):

Satrat (approved):

PR #412, opened 2024-01-22
More control over external data size

dbogunowicz:
As discussed recently:
- let's fold the splitting of big internal files into `onnx_save` 
- split the model.onnx into graph and constant tensors once the model.onnx is larger then DUMP_EXTERNAL_DATA_THRESHOLD (500MB)
- let's allow the user to control the size of created external tensor files 
 For usage see: https://github.com/neuralmagic/sparseml/pull/1983

### DIFF
--- a/src/sparsezoo/utils/onnx/external_data.py
+++ b/src/sparsezoo/utils/onnx/external_data.py
@@ -92,7 +92,10 @@ def save_onnx(
     :return True if the model was saved with external data, False otherwise.
     """
     if external_data_file is not None:
-        _LOGGER.debug(f"Saving with external data: {external_data_file}")
+        _LOGGER.debug(
+            f"Saving with external data, with file chunks of maximum size "
+            f"{MAX_FILE_SIZE_EXTERNAL_DATA / 1e6} MB"
+        )
         _check_for_old_external_data(
             model_path=model_path, external_data_file=external_data_file
         )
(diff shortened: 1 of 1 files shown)


mgoin:
@bfineran @markurtz this seems quite small, i would recommend at least a few gigabytes, if not the relatively standard 10GB used in HF

rahul-tuli:
I thought we were splitting the model into data files if it was >400mb and not restricting the model.data file(s) sizes? The only restriction on model.data file sizes is for easier download/upload from/to sparsezoo for which 400mb seems small? Thoughts?

bfineran (changes requested):
noting we shouldn't be updating chunk size here as discussed via slack - @dbogunowicz is aware of changes

dbogunowicz:
@alexm-nm just giving you the heads up, that splitting the external data is now part of the save_onnx function

src/sparsezoo/utils/onnx/external_data.py
```diff
@@ -98,13 +112,15 @@ def save_onnx(
...
         return True
 
-    if model.ByteSize() > onnx.checker.MAXIMUM_PROTOBUF:
+    if model.ByteSize() > DUMP_EXTERNAL_DATA_THRESHOLD:
         external_data_file = external_data_file or EXTERNAL_ONNX_DATA_NAME
         _LOGGER.warning(
             "The ONNX model is too large to be saved as a single protobuf. "
-            f"Saving with external data: {external_data_file}"
+            "Saving with external data, with file chunks of maximum size "
+            f"{DUMP_EXTERNAL_DATA_THRESHOLD / 1e9} GB"
```
bfineran, line 123: wrong threshold - also let's downgrade from warning - probably a bit to scary for an expected, default flow

src/sparsezoo/utils/onnx/external_data.py
```diff
@@ -98,13 +112,15 @@ def save_onnx(
             all_tensors_to_one_file=True,
             location=external_data_file,
         )
+        split_external_data(model_path, max_file_size=max_external_file_size)
         return True
 
-    if model.ByteSize() > onnx.checker.MAXIMUM_PROTOBUF:
+    if model.ByteSize() > DUMP_EXTERNAL_DATA_THRESHOLD:
         external_data_file = external_data_file or EXTERNAL_ONNX_DATA_NAME
         _LOGGER.warning(
```
dbogunowicz, line 120:
```suggestion
        _LOGGER.debug(
```

src/sparsezoo/utils/onnx/external_data.py
dbogunowicz, line 123:
```suggestion
            f"{max_external_file_size / 1e9} GB"
```

bfineran (approved):

PR #415, opened 2024-01-29
add support for benchmark.yaml

bfineran:
recent zoo models use `benchmark.yaml` instead of `benchmarks.yaml`. adding this additional pathway so `benchmark.yaml` is downloaded in the bulk model download

example:
```python
from sparsezoo import Model

stub = "zoo:llama2-7b-ultrachat200k_llama2_pretrain-base"

model = Model(stub)
model.benchmark.path
```

### DIFF
--- a/src/sparsezoo/model/utils.py
+++ b/src/sparsezoo/model/utils.py
@@ -58,6 +58,7 @@ ALLOWED_FILE_TYPES = {
     "benchmarking",
     "outputs",
     "onnx_gz",
+    "benchmark",
 }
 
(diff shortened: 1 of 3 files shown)


src/sparsezoo/model/utils.py
```diff
@@ -58,6 +58,7 @@
     "benchmarking",
     "outputs",
     "onnx_gz",
+    "benchmark",
```
rahul-tuli, line 61: Should `benchmarks` also be here?

rahul-tuli (approved):

PR #435, opened 2024-02-07
Download deployment dir for llms

rahul-tuli:
Fixes a bug for ModelAnalysis class, when using LLMs with external data files, using just model.onnx to load the model would fail

Fix is to download the deployment directory with external data files first and then load the model

### DIFF
--- a/src/sparsezoo/analyze/analysis.py
+++ b/src/sparsezoo/analyze/analysis.py
@@ -1305,6 +1305,7 @@ class ModelAnalysis(YAMLSerializableBaseModel):
             result = ModelAnalysis.from_onnx(Path(file_path) / "model.onnx")
 
         elif file_path.startswith("zoo:"):
+            Model(file_path).deployment.download()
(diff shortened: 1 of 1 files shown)


bfineran (approved):

src/sparsezoo/analyze/analysis.py
```diff
@@ -1305,6 +1305,7 @@ def create(cls, file_path: Union[str, ModelProto]) -> "ModelAnalysis":
             result = ModelAnalysis.from_onnx(Path(file_path) / "model.onnx")
 
         elif file_path.startswith("zoo:"):
+            Model(file_path).deployment.download()
```
bfineran, line 1308: Why not use .path to optionally download

rahul-tuli: Addressed! Good catch

mgoin (approved):
lgtm makes sense!

bfineran (approved):

PR #437, opened 2024-02-08
[sparsezoo.analyze] Fix pathway such that it works for larger models

dsikka:


src/sparsezoo/analyze_v2/model_analysis.py
```diff
@@ -146,20 +146,23 @@ def analyze(path: str, download_path: Optional[str] = None) -> "ModelAnalysis":
...
         model = Model(path, download_path)
         onnx_model_path = model.onnx_model.path
-        onnx_model = onnx.load(onnx_model_path)
+        onnx_model = onnx.load(onnx_model_path, load_external_data=False)
     else:
         raise ValueError(f"{path} is not a valid argument")
 
-    model_graph = ONNXGraph(onnx_model)
-    node_shapes, _ = extract_node_shapes_and_dtypes(model_graph.model)
+    node_shapes, _ = extract_node_shapes_and_dtypes(onnx_model, onnx_model_path)
```
bfineran, line 158: is the session cleaned up at this point so we aren't loading two copies of the full model in at once?

dsikka: should be - is there a way to explicitly make sure the inference session is cleaned up? deleting the object or explicitly running garbage collection?

seems like memory is released after the inference function completes/the session goes out of scope

src/sparsezoo/utils/onnx/external_data.py
```diff
@@ -174,19 +174,23 @@ def validate_onnx(model: Union[str, ModelProto]):
...
-def load_model(model: Union[str, ModelProto, Path]) -> ModelProto:
+def load_model(
+    model: Union[str, ModelProto, Path], load_external_data: bool = True
+) -> ModelProto:
     """
     Load an ONNX model from an onnx model file path. If a ModelProto
     is given, then it is returned.
 
     :param model: the model proto or path to the model ONNX file to check for loading
+    :param laod_external_data: if a path is given, whether or not to also load the
```
dbogunowicz, line 185:
```suggestion
    :param load_external_data: if a path is given, whether or not to also load the
```

dbogunowicz (dismissed):

bfineran (dismissed):

bfineran (approved):

horheynm (approved):

rahul-tuli (approved):

PR #451, opened 2024-02-15
[deepsparse.analyze] Fix v1 functionality to  work with llms

dsikka:
# Summary
- A sereis of changes were made to `sparsezoo.analyze` such that the functionality works with llms
- `deepsparse.analzye` is still using the sparsezoo analyze v1 pathway and shares different entry points compared to `v2` and `sparsezoo.analyze`
- This PR basically makes the same changes that were made for `sparsezoo.analyze` such that the v1 pathway also still works
- Ticket: https://app.asana.com/0/1206109050183159/1206512842320229/f

# Testing
- Local runs look good; still needs more testing

### DIFF
--- a/src/sparsezoo/analyze/analysis.py
+++ b/src/sparsezoo/analyze/analysis.py
@@ -1377,7 +1377,7 @@ class ModelAnalysis(YAMLSerializableBaseModel):
         if path:
             model = load_model(path)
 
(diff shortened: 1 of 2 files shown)


rahul-tuli (approved):
Nice! Thanks!

PR #470, opened 2024-02-28
Download iso chunk folder

horheynm:
Previous commit fails bc the chunks are saved in root model folder. 
Tests checks the expected files, where chunks are not included. 

Before these commits, chunk folder is deleted each run so it passed. Now if multiple downloading is running, then delete of a folder will have directory to write to

Chunks are now saved in `~/.cache/sparsezoo/neuralmagic/chunks/{stub}/{training_or_deployment_or_folder_name}/{model_data}/chunk1`

Combined chunks are saved in 
Chunks are now saved in `~/.cache/sparsezoo/neuralmagic/{stub}/training/model.data`

### DIFF
--- a/src/sparsezoo/utils/download.py
+++ b/src/sparsezoo/utils/download.py
@@ -93,13 +93,24 @@ class Downloader:
         self.chunk_bytes = chunk_bytes
         self.job_queues = Queue()
         self._lock = threading.Lock()
-        self.chunk_folder = self.get_parent_chunk_folder(download_path)
+        self.chunk_download_path = self.get_chunk_download_path(download_path)
 
-    def get_parent_chunk_folder(self, path: str) -> str:
-        """Get the name of the file that is used as folder inside chunks"""
-        path = path.split(os.path.sep)[-1]
-        path = path.replace(".", "_")
-        return path
+    def get_chunk_download_path(self, path: str) -> str:
+        """Get the path where chunks will be downloaded"""
+
+        # make the folder name from the model name and file to be downloaded
+        stub = path.split(os.path.sep)[-3]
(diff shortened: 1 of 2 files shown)


src/sparsezoo/utils/download.py
```diff
@@ -93,13 +93,24 @@ def __init__(
...
+
+        # make the folder name from the model name and file to be downloaded
+        stub = path.split(os.path.sep)[-3]
+        path = "_".join(path.split(os.path.sep)[-2:])
+        file_name_as_folder = path.replace(".", "_")
+
+        # save the chunks on a different folder than the root model folder
+        return os.path.join(
+            os.path.abspath(
+                os.path.expanduser("~/.cache/sparsezoo/neuralmagic/chunks")
```
bfineran, line 109: Can we get this from a constant elsewhere? ~ is going to be OS dependent I believe

bfineran (approved):

src/sparsezoo/utils/download.py
```diff
@@ -105,9 +106,8 @@ def get_chunk_download_path(self, path: str) -> str:
 
         # save the chunks on a different folder than the root model folder
         return os.path.join(
-            os.path.abspath(
-                os.path.expanduser("~/.cache/sparsezoo/neuralmagic/chunks")
-            ),
+            str(Path.home()),
+            "chunks",
```
bfineran, line 110: where did `.cache/sparsezoo/neuralmagic` go?

horheynm: Oh yes very good catch

bfineran (approved):

PR #473, opened 2024-03-04
Add build workflows for dev, nightly, and release wheels

dsikka:
# Summary
- Add a workflow which based on the conditions, will create a nightly, dev or release wheel
- The workflows make use of the actions in `nm-actions` and relies on this PR: https://github.com/neuralmagic/nm-actions/pull/3
- The following cases are covered:

## 1. Dev
- For dev, a wheel is created if a PR to main is opened or updated. This wheel will be called `sparsezoo_dev-[version].dev[PR]` where [PR] will be replaced to use the PR number and [version] will be replaced to the current version listed in `version.py`. All dev PRs are pushed to the internal bucket in s3 (from which jenkins should push to the internal PyPi server)

## 2. Release
- When a PR is opened/updated to a release branch: in that case, the built wheel will be pushed to the internal bucket in s3 and the wheel will be called `sparsezoo-[version]` where the [version] will be replaced to the current version listed in `version.py` 
- When a new release tag is cut, the naming will be the same however, the release will be pushed to the external/production s3 bucket for the production PyPi server

## 3. Nightly
- When a PR is merged into main, the wheel will be built and named `sparsezoo_nightly-[version].[date]`
- A scheduled nightly cron job will trigger a nightly build and create a wheel with the same name. The wheel will be pushed to an external/production s3 bucket for the production PyPi server

The equivalent workflow can be added in to sparseml once approved

# Testing
- Nightly and dev workflows have been tested however, this is with a dummy s3 bucket so we do not overload the buckets being used by the current workflow
- Release builds have not been tested 

# NOTE:
- Before this PR can be merged, the actions PR mentioned above must be merged

# Next Steps:
- Update jenkins to use the new/updated buckets
- Find a non-destructive way to test release workflows

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -4,7 +4,7 @@ on:
     types: [opened, synchronize, reopened]
     branches:
       - main
-      - 'release/**'
+      - 'release/[0-9]+.[0-9]+'
   push:
     branches:
       - main
@@ -26,7 +26,7 @@ concurrency:
 # missing release tag workflow/needs to be added in
 env:
   INTERNAL: ${{ github.event_name != 'schedule' && github.event_name != 'release'}}
-  RELEASE: ${{ github.event_name =='release' || (github.base_ref == 'release/**' && github.event_name == 'pull_request')}}
+  RELEASE: ${{ github.event_name =='release' || (startsWith(github.base_ref, 'release/') && github.event_name == 'pull_request')}}
   DEV: ${{ github.base_ref == 'main' && github.event_name == 'pull_request'}}
   NAME: ${{ github.event.number }} 
 
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -45,14 +45,14 @@ jobs:
           aws-region: us-east-1 
       - name: Build PyPi Wheel
         id: build-wheel
-        uses: neuralmagic/nm-actions/actions/pypi_build@pypi-actions
+        uses: neuralmagic/nm-actions/actions/pypi_build
         with:
           dev: $DEV
           release: $RELEASE
           name: $NAME
       - name: Push to s3 bucket
         id: push-wheel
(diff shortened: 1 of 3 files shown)


bfineran (dismissed):
looks great especially with the usage of the actions

horheynm (dismissed):

dsikka:
keeping open until relevant nm-actions are merged

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,58 @@
...
+    outputs:
+      wheel: ${{ steps.push-wheel.outputs.wheel }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Login to s3
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ secrets.AWS_WEBIDENTITY_FOR_GITHUB_ACTIONS }}
+          aws-region: us-east-1
```
horheynm, line 45: I think right now everything is us-east-1, so its ok. 
As we scale we might need to change this

src/sparsezoo/version.py
```diff
@@ -22,14 +22,17 @@
...
-        version_base
-        if is_release
-        else f"{version_base}.{date.today().strftime('%Y%m%d')}"
-    )
+    if is_release:
+        return version_base
+    elif is_dev:
+        return f"{version_base}.dev{dev_number}"
+    else:
+        return f"{version_base}.{date.today().strftime('%Y%m%d')}"
```
horheynm, line 35: This code looks fine bc the sed stuff got merged. Why did we stick with the sed command and not setting env vars for checking is_dev, is_.*?
The problem with using sed is that unless you know that the GHA exists, you wont know when they will be true. 
If using sed, Id add comment saying these variables will be overwritten from GHA

dsikka: sed is setting the values for the flags (i.e setting is_dev, is_version). And then the flags are set here.

horheynm (approved):

PR #474, opened 2024-03-05
Retire `flaky` 

dbogunowicz:
Remove the dependency on `flake` (which is not actively maintained)


mgoin (approved):
Is it not used in this repo?

bfineran (approved):

PR #476, opened 2024-03-08
Multiple download bug

xkazuka:
Ticket:
https://app.asana.com/0/1206109050183159/1206698210421769

Fix for multiple processing running the download code. 
Problem with the previous code is that once the download happens, it generates a folder name by the filename in the temp file and saves the chunk files there. If multiple download of the same file happens, the downloading folder can be deleted while downloading. 
Solution is to make a file_id for each download call and save it in there before going into the original download folder path.


rahul-tuli:
Is there a easy way to unit test this?

rahul-tuli (approved):

horheynm:
> Is there a easy way to unit test this?

Yes there is, ran fsdp where gpus can download the same files. 
https://app.asana.com/0/1206109050183159/1206698210421769

bfineran (approved):

PR #483, opened 2024-04-18
Update pydantic support

rahul-tuli:
This PR upgrades pydantic support from v1 to v2

The changes include:
- Using pydantic's `bump-pydantic` tool for the initial migration
- Updating typing of class vars
- Left FeatureStatus table to use BaseModel from `pydantic.v1` (shipped as a part of v2) as it is a low importance feature

Test method:
- Ran make test locally all tests are green

```bash
===================== 345 passed, 126 warnings in 656.59s (0:10:56) ======================
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ _deps = [
     "pyyaml>=5.1.0",
     "requests>=2.0.0",
     "tqdm>=4.0.0",
-    "pydantic>=2.0.0,<2.7",
+    "pydantic>=2.0.0,<2.8",
     "click>=7.1.2,!=8.0.0",  # latest version < 8.0 + blocked version with reported bug
     "protobuf>=3.12.2",
     "pandas>1.3",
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -51,7 +51,7 @@
     "pyyaml>=5.1.0",
     "requests>=2.0.0",
     "tqdm>=4.0.0",
-    "pydantic>=1.8.2,<2.0.0",
```
dbogunowicz, line 54: Please delete : local/features/pydantic-upgrade/v2_validation.py

dbogunowicz (dismissed):

setup.py
rahul-tuli: Oof good catch Damian, I thought this folder was in my local .gitignore

dbogunowicz (approved):

rahul-tuli:
Note: We should wait until sparseml pydantic upgrade is ready to land

bfineran (approved):

